PR #4, opened 2024-04-21
Make assistant tokenization error message more specific

patrickvonplaten:
Most tokenizer exception arguably come from the message being empty. Currently this gives the following error message:
```sh
'Invalid assistant message: '
```

which looks a bit incomplete. I think it'd be better to throw:
```sh
'Empty assistant message.'
```


src/mistral_common/tokens/tokenizers/sentencepiece.py
```diff
@@ -289,6 +289,8 @@ def encode_assistant_message(self, message: AssistantMessage, is_before_last_use
             ]
         elif message.content:
             curr_tokens = self.tokenizer.encode(message.content, bos=False, eos=False)
+        elif message.content == "":
+            raise TokenizerException(f"Empty assistant message.")
```
patrickvonplaten:
```suggestion
            raise TokenizerException("Empty assistant message.")
```

pandora-s-git (approved):

PR #5, opened 2024-04-24
Allow multiple "assistant" messages to follow each other

abellion:
This PR makes it possible to send a set of messages within which two consecutive messages have the "assistant" role (it currently does not pass the validation). Doing so streamlines the validation in regards to the other roles (they can all be followed by another message of the same role), and also restores the behaviour of the API as of before April 17.


juliendenize:
This PR is outdated, already done upstream.

PR #21, opened 2024-06-08
typo: Correct "recieved" to "received" in MistralRequestValidator

CharlesCNorton:
Corrected a typo in the MistralRequestValidator class. The word "recieved" was misspelled and has been changed to "received" to ensure proper spelling in the comment "if we have an assistant message and we have not received all the function calls".


SCAuFish:
Thanks Charles for the fixes!
I would suggest combining the few typo-fix PRs into one, so that they are easier to review, as they don't require as much modularization as code change? 
I am new to this repo, though, so open to comments from more active contributors.

patrickvonplaten (approved):

PR #30, opened 2024-07-18
tests: add tekken instruct v3

jean-malo:


poetry.lock
```diff
@@ -33,6 +33,17 @@ tests = ["attrs[tests-no-zope]", "zope-interface"]
 tests-mypy = ["mypy (>=1.6)", "pytest-mypy-plugins"]
 tests-no-zope = ["attrs[tests-mypy]", "cloudpickle", "hypothesis", "pympler", "pytest (>=4.3.0)", "pytest-xdist[psutil]"]
 
+[[package]]
```
patrickvonplaten, line 36: Are all these updates really needed?

tests/test_tokenize_v3.py
```diff
@@ -205,15 +323,37 @@ def test_tool_message_multiple_shots_with_history(tokenizer: InstructTokenizer)
...
+        (
+            tokenizer(),
+            "<s>[INST]▁a[/INST]"
+            '[TOOL_CALLS]▁[{"name":▁"b",▁"arguments":▁{},▁"id":▁"0"},▁{"name":▁"q",▁"arguments":▁{},▁"id":▁"1"}]</s>'
+            '[TOOL_RESULTS]▁{"content":▁"d",▁"call_id":▁"0"}[/TOOL_RESULTS]'
+            '[TOOL_RESULTS]▁{"content":▁"d",▁"call_id":▁"1"}[/TOOL_RESULTS]'
+            "▁e</s>[INST]▁f[/INST]"
+            '[TOOL_CALLS]▁[{"name":▁"b",▁"arguments":▁{},▁"id":▁"2"},▁{"name":▁"q",▁"arguments":▁{},▁"id":▁"3"}]</s>'
+            '[TOOL_RESULTS]▁{"content":▁"d",▁"call_id":▁"2"}[/TOOL_RESULTS]'
+            '[TOOL_RESULTS]▁{"content":▁"d",▁"call_id":▁"3"}[/TOOL_RESULTS]',
```
patrickvonplaten: (nit) could we maybe define these templates above:

```py
TEST_TOOL_TEMPLATES = ...

```

it could make it easier to read and maintain

poetry.lock
jean-malo: This commit didn't lock the files
https://github.com/mistralai/mistral-common/commit/d991efa48285b80cbe8c3b1b84043826732c60e4

patrickvonplaten (approved):
Very nice tests! Quite some boilerplate code in the tests (maybe we could use global variables instead?). 

Thanks a bunch for adding it!

jean-malo:
> Very nice tests! Quite some boilerplate code in the tests (maybe we could use global variables instead?).

We could but they have slight variations (eg. tekken doesn't have ` ` between special tokens but keeps the space in regular text while sentencepiece has `_` everywhere. So it's a bit annoying and the templates would look quite convoluted, I think repetition is ok in tests as it makes debugging slightly easier

jean-malo:
If you have a simple template example, happy to implement it but the ones I came up with just made it worse

patrickvonplaten:
> > Very nice tests! Quite some boilerplate code in the tests (maybe we could use global variables instead?).
> 
> We could but they have slight variations (eg. tekken doesn't have ` ` between special tokens but keeps the space in regular text while sentencepiece has `_` everywhere. So it's a bit annoying and the templates would look quite convoluted, I think repetition is ok in tests as it makes debugging slightly easier

Guess template would nicely show the differences (maybe just as follows):

```py
FN_CALL_TEMPLATE = '[TOOL_RESULTS]▁{"content":{ }"d",{ }"call_id"{ }"3"}[/TOOL_RESULTS]',
```

and then later:

```py
FN_CALL_TEMPLATE.format(...)
```

But also ok for me to leave as is

jean-malo:
@patrickvonplaten added some form of template, lmk what you think

Issue #38, opened 2024-08-13
[BUG]: Upgrade pydantic dep to `^2.7.4` to allow compatibility with Python 2.12


egeres:
### Python -VV

```shell
Python 3.12.4 | packaged by Anaconda, Inc. | (main, Jun 18 2024, 15:12:24) [GCC 11.2.0]
```

### Pip Freeze

```shell
aiohttp==3.9.5
aiosignal==1.3.1
annotated-types==0.7.0
anyio==4.4.0
attrs==23.2.0
certifi==2024.7.4
charset-normalizer==3.3.2
cowsay==6.1
distro==1.9.0
frozenlist==1.4.1
greenlet==3.0.3
h11==0.14.0
httpcore==1.0.5
httpx==0.27.0
idna==3.7
iniconfig==2.0.0
jsonpatch==1.33
jsonpointer==3.0.0
jsonschema==4.21.1
jsonschema-specifications==2023.12.1
langchain==0.2.5
langchain-core==0.2.8
langchain-openai==0.1.14
langchain-text-splitters==0.2.1
langsmith==0.1.81
loguru==0.7.2
markdown-it-py==3.0.0
mdurl==0.1.2
mistral_common==1.3.3
multidict==6.0.5
numpy==1.26.4
openai==1.37.1
orjson==3.10.6
packaging==24.1
pluggy==1.5.0
pydantic==2.6.1
pydantic_core==2.16.2
Pygments==2.18.0
pytest==8.3.2
PyYAML==6.0.1
referencing==0.35.1
regex==2024.7.24
requests==2.32.3
rich==13.7.1
rpds-py==0.20.0
ruff==0.5.5
sentencepiece==0.2.0
setuptools==69.5.1
sniffio==1.3.1
SQLAlchemy==2.0.31
tenacity==8.5.0
tiktoken==0.7.0
tqdm==4.66.4
typing_extensions==4.12.2
urllib3==2.2.2
wheel==0.43.0
yarl==1.9.4
```

### Reproduction Steps

Run:

```python
from langchain_core.output_parsers.base import BaseOutputParser
```

### Expected Behavior

That single import should work, but it instead raises the following error:

```
TypeError: ForwardRef._evaluate() missing 1 required keyword-only argument: 'recursive_guard'
```

### Suggested Solutions

[...]

PR #39, opened 2024-08-13
Updated Pydantic version to `^2.8.2`

egeres:
More information on: https://github.com/mistralai/mistral-common/issues/38

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ ignore_missing_imports = true
 
 [tool.poetry.dependencies]
 python = "^3.8.10"
-pydantic = "2.6.1"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -28,7 +28,7 @@ ignore_missing_imports = true
 
 [tool.poetry.dependencies]
 python = "^3.8.10"
-pydantic = "2.6.1"
+pydantic = "^2.8.2"
```
patrickvonplaten, line 31:
```suggestion
pydantic = "^2.6.1"
```

This should be fine as well no?

patrickvonplaten: It'll by default install the newest version of pydantic then!

PR #51, opened 2024-09-18
Add headers to download images

ywang96:
Some image urls (e.g., wikipedia images) requires a policy for the requester to have an explicit user-agent header, and will return `403 Client Error: Forbidden` if it's missing.

This PR adds the header to the `download_image` function. Currently we use wikipedia images [for the example of the vision API](https://docs.vllm.ai/en/latest/models/vlm.html#openai-vision-api) in our documentation on vLLM, so it would be great if our users can run this example for `mistralai/Pixtral-12B-2409` too.

### DIFF
--- a/src/mistral_common/multimodal.py
+++ b/src/mistral_common/multimodal.py
@@ -7,11 +7,13 @@ from PIL import Image
 from pydantic import BeforeValidator, PlainSerializer, SerializationInfo
 from typing_extensions import Annotated
 
+from mistral_common import __version__
 
 def download_image(url: str) -> Image.Image:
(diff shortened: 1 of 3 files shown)


ywang96:
cc @patrickvonplaten Let me know if this is the best way to do this! We have [something similar](https://github.com/vllm-project/vllm/blob/95965d31b6ac2c9557816a6ffabe4a3117a5ccb2/vllm/connections.py#L43) in vLLM, but open to change the header content to anything else!

patrickvonplaten (approved):
Thanks a lot for fixing this @ywang96 - just commited a quick test to your PR. 

Will proceed with a patch release now

PR #58, opened 2024-09-28
[cv2] Make sure broken environments don't lead to errors

patrickvonplaten:
Sadly it often happens that cv2 is incorrectly installed: https://github.com/opencv/opencv-python/issues/884 which shouldn't bother people from using mistral_common's text encoding capabilities. 

This PR makes sure that all import errors are caught and instead of raising we simply throw a warning.

cc @youkaichao as well


timlacroix (approved):

youkaichao:
thanks for the update!

src/mistral_common/tokens/tokenizers/multimodal.py
```diff
@@ -14,13 +15,24 @@
...
 _cv2_installed: bool
 try:
     import cv2
 
     _cv2_installed = True
 except ImportError:
     _cv2_installed = False
+except Exception as e:
+    # cv2 has lots of import problems: https://github.com/opencv/opencv-python/issues/884
+    # for better UX, let's simple skip import errors for now
```
patrickvonplaten:
```suggestion
    # for better UX, let's simply skip all errors that might arise from import for now
```

PR #60, opened 2024-10-17
Fix exceptions when constructing an exception messages

sasha0552:
Because of the `use_enum_values=True` defined in the base Pydantic model, enum fields are parsed as strings instead of enum values. This causes exceptions like `AttributeError: 'str' object has no attribute 'value'` when constructing the actual exception message.

https://github.com/mistralai/mistral-common/blob/4a1d7c2d645242721c81e689d46cf0a86582a507/src/mistral_common/base.py#L9

This PR fixes this.

### DIFF
--- a/src/mistral_common/protocol/instruct/validator.py
+++ b/src/mistral_common/protocol/instruct/validator.py
@@ -217,7 +217,7 @@ class MistralRequestValidator(Generic[UserMessageType, AssistantMessageType, Too
 
                 if current_role not in expected_roles:
                     raise InvalidMessageStructureException(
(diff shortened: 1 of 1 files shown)


sasha0552:
@patrickvonplaten Could you please review & consider merging? Without these fixes, vLLM returns unhelpful messages to users (`'str' object has no attribute 'value'` without further explanation) when using the Mistral tokenizer and messing with the order of roles.

patrickvonplaten:
Hey @sasha0552, 

thanks for the fixes - this looks indeed good to me!

patrickvonplaten (approved):

PR #77, opened 2025-03-16
Add spatial merge size

patrickvonplaten:
This PR adds spatial merge size


tests/test_multimodal.py
```diff
@@ -49,7 +52,11 @@ def test_download_gated_image(mm_config: MultimodalConfig, special_token_ids: Sp
         assert image is not None, "Make sure gated wikipedia images can be downloaded"
 
 
-def test_image_encoder(mm_config: MultimodalConfig, special_token_ids: SpecialImageIDs) -> None:
+@pytest.mark.parametrize("spatial_merge_size", [1, 2])
+def test_image_encoder(spatial_merge_size: int, special_token_ids: SpecialImageIDs) -> None:
+    mm_config = MultimodalConfig(
+        image_patch_size=16 // spatial_merge_size, max_image_size=128, spatial_merge_size=spatial_merge_size
```
timlacroix, line 58: why are we dividing here ? I'm surprised.
If we want to keep all numbers of tokens the same shouldn't we multiply ? 

I assume this is correct though otherwise we'd have shit results...

patrickvonplaten: Hmm maybe the test is misleading - when constructing a `mm_config` we should not pass `image_patch_size // spatial_merge_size`, but just `image_patch_size`.

However, we know that the two following configs are identical from the point of view of the tokenizer:
`image_patch_size=16, spatial_merge_size=1`
and:
`image_patch_size=8, spatial_merge_size=2`

=> Think equality tests are always quite powerful that's why it's done here

patrickvonplaten: So TLDR is that it's just for testing

timlacroix (approved):
lgtm

PR #78, opened 2025-04-09
Remove extra="forbid" from the base pydantic model

tauven:
Remove the extra="forbid" so extra values are ignored.
Extra values should be ignored for better compatibility and to allow usage with other tools. (for example: serving models with vllm)

see: https://github.com/vllm-project/vllm/issues/12864


ItzAmirreza:
Thanks for the fix, looking forward to the merge.

tauven:
The issue should be resolved with the changes made inhttps://github.com/vllm-project/vllm/pull/20973

PR #82, opened 2025-05-22
[Tokenizer >= v7] Allow content and tool calls to be defined together

patrickvonplaten:
Allow content and tool calls to be jointly defined

### DIFF
--- a/src/mistral_common/protocol/instruct/normalize.py
+++ b/src/mistral_common/protocol/instruct/normalize.py
@@ -35,8 +35,8 @@ class InstructRequestNormalizer(
     - Normalize tool calls
     """
 
-    system_prompt_in_begin: bool = False
-    allow_tool_call_and_content: bool = False
+    _system_prompt_in_begin: bool = False
+    _allow_tool_call_and_content: bool = False
 
     def __init__(
         self,
@@ -121,7 +121,7 @@ class InstructRequestNormalizer(
         for message in messages:
             assert isinstance(message, self._assistant_message_class), "Expected assistant message"
 
-            if not self.allow_tool_call_and_content and (message.tool_calls and message.content):
+            if not self._allow_tool_call_and_content and (message.tool_calls and message.content):
                 raise ValueError(f"Tool calls and content cannot be used together in the same message. {message}")
 
             if message.tool_calls:
@@ -215,7 +215,7 @@ class InstructRequestNormalizer(
         # If the first message is not a user message, or we didnt aggregate
         # anything (all system messages) for example, add an empty user message
         if len(aggregated_messages) == 0 or (
-            not self.system_prompt_in_begin and aggregated_messages[0].role != Roles.user
+            not self._system_prompt_in_begin and aggregated_messages[0].role != Roles.user
         ):
             aggregated_messages.insert(0, self._user_message_class(content=""))
 
@@ -231,8 +231,8 @@ class InstructRequestNormalizer(
 
 
 class InstructRequestNormalizerV7(InstructRequestNormalizer):
-    system_prompt_in_begin: bool = True
-    allow_tool_call_and_content: bool = True
+    _system_prompt_in_begin: bool = True
+    _allow_tool_call_and_content: bool = True
 
     @staticmethod
     def normalizer() -> "InstructRequestNormalizerV7":
--- a/src/mistral_common/protocol/instruct/normalize.py
+++ b/src/mistral_common/protocol/instruct/normalize.py
@@ -34,6 +34,7 @@ class InstructRequestNormalizer(
     - Normalize json content
     - Normalize tool calls
     """
+
     system_prompt_in_begin: bool = False
     allow_tool_call_and_content: bool = False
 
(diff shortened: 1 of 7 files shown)


src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -295,6 +295,9 @@ def _encode_tool_calls_in_assistant_message(self, message: AssistantMessageType)
         return curr_tokens
 
     def encode_assistant_message(self, message: AssistantMessageType, is_before_last_user_message: bool) -> List[int]:
+        if message.tool_calls and message.content:
```
joa-stdn, line 298: I think to be exhaustive maybe we should also raise in `InstructTokenizerV1` that defines its own `encode_assistant_message`? (no need for V3 that calls `super()`)

tests/test_normalization.py
```diff
@@ -339,6 +306,112 @@ def test_normalize_funcalls(self, normalizer: InstructRequestNormalizer) -> None
...
+        assert isinstance(first_message, SystemMessage)
+        assert first_message.content == "S"
+
+        second_message = parsed_request.messages[1]
+        assert isinstance(second_message, AssistantMessage)
+        assert second_message.content == "A"
+
+        assert parsed_request.system_prompt is None
+
+    def test_assistant_assistant_system_v7(self, normalizer_v7: InstructRequestNormalizer) -> None:
```
joa-stdn, line 335: The name of this test is not adapted?

tests/test_normalization.py
```diff
@@ -339,6 +306,112 @@ def test_normalize_funcalls(self, normalizer: InstructRequestNormalizer) -> None
...
+        chat_completion_request = mock_chat_completion(
+            messages=[
+                UserMessage(content="A1"),
+                AssistantMessage(
+                    content="B1",
+                ),
+                AssistantMessage(
+                    content="B2",
+                    tool_calls=[ToolCall(function=FunctionCall(name="tool1", arguments='{"input": "1"}'))],
+                ),
```
joa-stdn, line 384: Not sure this example makes a lot of sense in practice because we're missing the tool result messages in between? I would say that if anything this example is not that great since it's messing up with the order in tokenization and hiding the fact that B2 led to tool1 and B4 to tools tool21 and tool22
Should we even allow successive assistant messages without tool results?

joa-stdn: Besides I think we should add another test with normalizer v7 composed with tokenizer v7 to be exhaustive

joa-stdn (commented):
left a few questions

src/mistral_common/tokens/tokenizers/instruct.py
patrickvonplaten: Hmm `InstructTokenizerV1` doesn't allow tool calls at all so think we're good here

tests/test_normalization.py
patrickvonplaten: I just moved it from above

tests/test_normalization.py
patrickvonplaten: Good idea with end-to-end test. 

Think the test chat completion request is a good example though because we want to keep it as minimal as possible and because we can have multiple assistant messages that will be aggregated together

src/mistral_common/protocol/instruct/normalize.py
```diff
@@ -36,6 +36,7 @@ class InstructRequestNormalizer(
     """
 
     system_prompt_in_begin: bool = False
+    allow_tool_call_and_content: bool = False
```
juliendenize: I like that below you rename `allow_tool_call_and_content` with a leading `_` to mark it private. Can we also do that for these args ? The user is not supposed to handle this right ?

tests/test_tokenizer_v7.py
```diff
@@ -342,3 +347,49 @@ def test_from_model() -> None:
...
+    # make sure it also works end to end
+    tools = request.available_tools
+    chat_completion_request = ChatCompletionRequest(
+        **request.model_dump(exclude={"system_prompt", "truncate_at_max_tokens", "available_tools"}), tools=tools
+    )
+    validator = MistralRequestValidatorV5(mode=ValidationMode.finetuning)
+    normalizer = InstructRequestNormalizerV7.normalizer()
+
+    mistral_tokenizer = MistralTokenizer(tekkenizer, validator, normalizer)
+    mistral_tokenizer._chat_completion_request_validator
```
juliendenize: I think you forgot to test something here

juliendenize (approved):
LGTM, just some minor points feel free to merge if I'm wrong

tests/test_tokenizer_v7.py
juliendenize: by here I mean the L391

juliendenize: by here I mean the L391

src/mistral_common/protocol/instruct/normalize.py
patrickvonplaten: Yeah fair

tests/test_tokenizer_v7.py
patrickvonplaten: Thanks!

PR #85, opened 2025-06-05
Add support for file:// protocol URIs.

sjuxax:
vLLM supports `file://` urls, but it fails without this when using the Mistral tokenizer. You may wish to place additional security machinery around this; in vLLM, one must pass `--allowed-local-media-path` on the commandline.

### DIFF
--- a/src/mistral_common/tokens/tokenizers/multimodal.py
+++ b/src/mistral_common/tokens/tokenizers/multimodal.py
@@ -48,6 +48,8 @@ def image_from_chunk(chunk: Union[ImageURLChunk, ImageChunk]) -> SerializableIma
         data = chunk.get_url().split(",")[1]
         image_data = base64.b64decode(data)
         return Image.open(BytesIO(image_data))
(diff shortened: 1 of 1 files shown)


src/mistral_common/tokens/tokenizers/multimodal.py
```diff
@@ -48,6 +48,8 @@ def image_from_chunk(chunk: Union[ImageURLChunk, ImageChunk]) -> SerializableIma
         data = chunk.get_url().split(",")[1]
         image_data = base64.b64decode(data)
         return Image.open(BytesIO(image_data))
+    if chunk.get_url().startswith("file"):
+        return Image.open(open(chunk.get_url().replace("file://", ""),'rb'))
```
juliendenize:
```suggestion
        return Image.open(open(chunk.get_url().replace("file://", ""), "rb"))
```

juliendenize (approved):
Thanks !

PR #90, opened 2025-06-06
Add the from_hf_hub method to MistralTokenizer

juliendenize:
Add support to download our tokenizers from the HuggingFace hub directly inside `mistral-common`.

This would ease for users the usage of our tokenizers by preventing them to use a specific code snippet.

To access this method they need to install `mistral-common` like this:
```sh
pip install mistral-common[hf-hub]
```
Or install the hub package by themselves.


patrickvonplaten (commented):
Let's maybe wait a bit more with this one to see to what extent `mistral_common` can be integrated into `transformers`

PR #92, opened 2025-06-09
Add support to the openai format for Chat completions

juliendenize:
This PR aims to add support to the OpenAI format for chat completion.

It adds converter from and to an OpenAI format with the following considerations:
- We do not add the OpenAI dependency which means we export and import dictionaries. It's fine as OpenAI uses typing via `TypedDict`
- We handle some naming conversions between their and our format.
- Export of requests allow adding parameters from the OpenAI format (e.g `InstructRequest.to_openai(temperature=0.15)` works). We verify parameters name are correct but not their types.

The tests have the OpenAI dependency to ensure our conversion is correct.

### DIFF
--- a/src/mistral_common/protocol/instruct/request.py
+++ b/src/mistral_common/protocol/instruct/request.py
@@ -171,10 +171,10 @@ class ChatCompletionRequest(BaseCompletionRequest, Generic[ChatMessageType]):
         Returns:
             The chat completion request.
         """
-        if "seeds" in kwargs and "random_seed" in kwargs:
-            raise ValueError("Cannot specify both `seeds` and `random_seed`.")
+        if "seed" in kwargs and "random_seed" in kwargs:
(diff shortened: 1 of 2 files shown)


tests/test_converters.py
```diff
@@ -0,0 +1,559 @@
...
+from openai.types.chat.chat_completion_content_part_text_param import (
+    ChatCompletionContentPartTextParam as OpenAITextChunk,
+)
+from openai.types.chat.chat_completion_message_tool_call_param import (
+    ChatCompletionMessageToolCallParam as OpenAIToolCall,
+)
+from openai.types.chat.chat_completion_system_message_param import (
+    ChatCompletionSystemMessageParam as OpenAISystemMessage,
+)
+from openai.types.chat.chat_completion_tool_message_param import ChatCompletionToolMessageParam as OpenAIToolMessage
```
patrickvonplaten, line 22: Nice that works for me

patrickvonplaten (approved):

PR #102, opened 2025-06-17
Fix error message instructions in transform_image()

gaby:
The error message in `transform_image` indicates the user to install `mistral_common` instead of `mistral-common`

### DIFF
--- a/src/mistral_common/tokens/tokenizers/multimodal.py
+++ b/src/mistral_common/tokens/tokenizers/multimodal.py
@@ -141,7 +141,9 @@ def transform_image(image: Image.Image, new_size: Tuple[int, int]) -> np.ndarray
         Transformed image with shape (C, H, W).
     """
     if not is_cv2_installed():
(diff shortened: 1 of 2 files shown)


juliendenize (approved):
Thanks !

src/mistral_common/tokens/tokenizers/multimodal.py
```diff
@@ -141,7 +141,9 @@ def transform_image(image: Image.Image, new_size: Tuple[int, int]) -> np.ndarray
         Transformed image with shape (C, H, W).
     """
     if not is_cv2_installed():
-        raise ImportError("OpenCV is required for this function. Install it with 'pip install mistral_common[opencv]'")
+        raise ImportError(
+            "OpenCV is required for this function. Install it with 'pip install mistral-common[opencv]'"
+        )
```
juliendenize:
```suggestion
        raise ImportError("OpenCV is required for this function. Install it with 'pip install mistral-common[opencv]'")
```

juliendenize (approved):

PR #114, opened 2025-07-01
[Naming] Rename multi-modal to image

patrickvonplaten:


src/mistral_common/tokens/tokenizers/base.py
```diff
@@ -249,97 +247,23 @@ def file_path(self) -> Path:
 TokenizedType = TypeVar("TokenizedType", bound=Tokenized)
 
 
-@dataclass
```
patrickvonplaten, line 252: moved into image.py file

src/mistral_common/tokens/tokenizers/tekken.py
```diff
@@ -250,7 +250,14 @@ def from_file(cls: Type["Tekkenizer"], path: Union[str, Path]) -> "Tekkenizer":
         untyped["special_tokens"] = special_tokens
 
         if mm := untyped.get("multimodal", None):
-            untyped["multimodal"] = MultimodalConfig(**mm)
+            # TODO(Patrick) - enable warning once fixed in vllm & transformers
+            # warnings.warn(
+            #     "multimodal is deprecated. Please rename your tekken config to use image instead.",
+            #     FutureWarning,
+            # )
```
juliendenize: As we did for the special tokens, i do not think we should have a future warning. I believe that we should raise an error when the version will be > v11. Otherwise users will complain about receiving warnings that is not caused by them but by downloading tekken.json we provided on HF hub.

src/mistral_common/data/tekken_240911.json
```diff
@@ -750008,7 +750008,7 @@
             "token_str": " (\"*"
         }
     ],
-    "multimodal": {
```
juliendenize: Don't think we should update old tekkens

juliendenize: See comment in Tekkenizer for why I think that

src/mistral_common/data/tekken_240911.json
```diff
@@ -750012,4 +750012,4 @@
         "image_patch_size": 16,
         "max_image_size": 1024
     }
-}
\ No newline at end of file
+}
```
juliendenize, line 750015:
```suggestion
}

```

src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -30,24 +30,31 @@
...
-        self.mm_encoder = mm_encoder
-        super().__init__(tokenizer, mm_encoder)
+        self.image_encoder = image_encoder
+        super().__init__(tokenizer, image_encoder)
+
+    @property
+    def mm_encoder(self) -> Optional[ImageEncoder]:
+        # this funtion is deprecated, use image_encoder instead
+        # TODO(Patrick) - throw a deprecation warning once
+        # changes applied to vllm and transformers
```
juliendenize, line 56:
```suggestion
        # deprecated - only allowed for tokenizers <= v11
```

juliendenize (approved):
LGTM, gave just some minor modifications suggestions

src/mistral_common/tokens/tokenizers/instruct.py
patrickvonplaten: Hmm no this is code change that's independent from tokenizer files -> this should be deprecated

juliendenize: As discussed in Slack, you're right, resolving the convo !

PR #115, opened 2025-07-03
Add v13 tokenizer

juliendenize:
Add v13 tokenizer:
- no longer tokenize call ids
- reorder tool messages

Also now all types of messages accept chunks.

### DIFF
--- a/src/mistral_common/protocol/instruct/messages.py
+++ b/src/mistral_common/protocol/instruct/messages.py
@@ -378,7 +378,7 @@ class ToolMessage(BaseMessage):
        >>> message = ToolMessage(content="Hello, how can I help you?", tool_call_id="123")
     """
 
-    content: str
+    content: Union[str, List[ContentChunk]]
     role: Literal[Roles.tool] = Roles.tool
     tool_call_id: Optional[str] = None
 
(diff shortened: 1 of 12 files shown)


src/mistral_common/protocol/instruct/messages.py
```diff
@@ -314,7 +314,7 @@ class AssistantMessage(BaseMessage):
     """
 
     role: Literal[Roles.assistant] = Roles.assistant
-    content: Optional[str] = None
+    content: Optional[Union[str, List[ContentChunk]]] = None
```
patrickvonplaten, line 317: why do we allow non `str` content for the assistant? I don't think this is necessary here

src/mistral_common/protocol/instruct/messages.py
```diff
@@ -378,7 +387,7 @@ class ToolMessage(BaseMessage):
        >>> message = ToolMessage(content="Hello, how can I help you?", tool_call_id="123")
     """
 
-    content: str
+    content: Union[str, List[ContentChunk]]
```
patrickvonplaten, line 390: how can the tool message have a non `str` content?

PR #116, opened 2025-07-03
Add v13 Tokenizer

juliendenize:
Add v13 tokenizer:

- no longer tokenize call ids
- reorder tool messages

### DIFF
--- a/src/mistral_common/tokens/tokenizers/instruct.py
+++ b/src/mistral_common/tokens/tokenizers/instruct.py
@@ -238,7 +238,6 @@ class InstructTokenizerV1(
         Returns:
             The encoded tokens and empty list.
         """
-        assert message.content is not None
         assert isinstance(message.content, str), "Message content must be normalized"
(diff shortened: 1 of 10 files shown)


src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -816,12 +830,7 @@ def encode_assistant_message(
 
         curr_tokens: list = []
         if message.content:
-            if isinstance(message.content, str):
-                curr_tokens += self._encode_normal_content_assistant_message(message)
-            elif isinstance(message.content, list):
-                curr_tokens += self.encode_content_chunks(
-                    message.content, is_last=False, system_prompt=None, force_img_first=True
-                ).tokens
+            curr_tokens += self._encode_normal_content_assistant_message(message)
```
joa-stdn, line 827: Nice catch! Should we also remove `List[ContentChunk]` in content possibilities in system messages as well?

juliendenize: Should be done now

PR #120, opened 2025-07-15
Add AudioURLChunk

juliendenize:
Add supports to the AudioURLChunk to ensure compatibility with vLLM.

Also fixes Voxtral links in doc.


src/mistral_common/audio.py
```diff
@@ -93,6 +95,35 @@ def duration(self) -> float:
...
+            An instance of the Audio class.
+        """
+        try:
+            return Audio.from_file(url_or_string, strict=strict)
+        except (FileNotFoundError, OSError):
+            pass
+
+        try:
+            response = requests.get(url_or_string)
+            response.raise_for_status()
```
juliendenize: This is untested didn't find a correct link to simulate this.

src/mistral_common/audio.py
```diff
@@ -93,6 +95,35 @@ def duration(self) -> float:
...
+        r"""Create an Audio instance from a URL or a base64 encoded string.
+
+        Args:
+            url_or_string: The URL or base64 encoded string of the audio file.
+            strict: Whether to strictly enforce mono audio.
+
+        Returns:
+            An instance of the Audio class.
+        """
+        try:
```
patrickvonplaten, line 109: not a big fan of all the try ... except -> can we maybe instead just have individual `from_url(....)`? We already hav a `from_base64(...)` no?

patrickvonplaten: I think here we should only do `requests.get(url) => Audio.from_bytes(...)`

src/mistral_common/protocol/instruct/messages.py
```diff
@@ -161,28 +164,64 @@ class RawAudio(MistralBase):
     format: str
 
     @classmethod
-    def from_audio(cls, audio: Audio) -> "RawAudio":
+    def from_audio(cls, audio: Audio, prefix: bool = False) -> "RawAudio":
```
patrickvonplaten: Hmm I don't think we should add a prefix here

patrickvonplaten: I don't see a use case where we'd want to have a prefix as part of the "raw" audio

src/mistral_common/tokens/tokenizers/audio.py
```diff
@@ -157,8 +157,13 @@ def next_multiple_of_chunk_frames(self, audio_array_len: int, sampling_rate: int
 
         return math.ceil(audio_array_len / self.audio_config.chunk_frames) * self.audio_config.chunk_frames
 
-    def _encode_audio_chunk(self, content: AudioChunk) -> AudioEncoding:
-        audio = Audio.from_raw_audio(content.input_audio)
+    def _encode_audio_chunk(self, content: Union[AudioChunk, AudioURLChunk]) -> AudioEncoding:
+        if isinstance(content, AudioURLChunk):
```
patrickvonplaten: nice! But could we maybe just do `Audio.from_audio_url(content.audio_url)` and then have a couple if-else statements in it? This is a bit hacky here IMO and not fully in line with `from_raw_audio(...)`

juliendenize:
> Very nice! Cool that it all works. Can we try to remove some of the try...except statements and have "cleaner" `from...` methods?

Tried to move them a bit but still add to handle some cases :/

src/mistral_common/tokens/tokenizers/audio.py
```diff
@@ -1,15 +1,28 @@
...
 
 logger = logging.getLogger(__name__)
 
+_URL_SCHEMES = {
+    "http",
+    "https",
+    "ftp",
+    "s3",  # AWS S3
+    "gs",  # Google Cloud Storage
+    "azure",  # e.g. azure://container/blob
```
```suggestion
    "azure", 
```

src/mistral_common/tokens/tokenizers/audio.py
```diff
@@ -1,15 +1,28 @@
...
 
+_URL_SCHEMES = {
+    "http",
+    "https",
+    "ftp",
+    "s3",  # AWS S3
+    "gs",  # Google Cloud Storage
+    "azure", 
+    "wasbs",
+    "abfs",  # Azure Blob/ADLS
```
```suggestion
_URL_SCHEMES = {
    "http",
    "https",
    "ftp",
    "s3",
    "gs",
    "azure",
    "wasbs",
    "abfs",
```

src/mistral_common/audio.py
```diff
@@ -93,6 +95,24 @@ def duration(self) -> float:
...
+            response = requests.get(url)
+            response.raise_for_status()
+            return Audio.from_bytes(response.content, strict=strict)
+        except Exception as e:
+            raise ValueError(f"Either the URL is not valid or decoding failed: {url}") from e
```
patrickvonplaten:
```suggestion
            response.raise_for_status()
       except requests.exceptions.RequestException as err:
            raise ValueError(f"Either the URL is not valid or decoding failed: {url}") from e
            
      return Audio.from_bytes(response.content, strict=strict)
```

can we catch more specifically here?

src/mistral_common/protocol/instruct/messages.py
```diff
@@ -171,18 +174,52 @@ def from_audio(cls, audio: Audio) -> "RawAudio":
...
+
+    type: Literal[ChunkTypes.audio_url] = ChunkTypes.audio_url
+    audio_url: Union[str, AudioURL]
+
+    def to_openai(self) -> Dict[str, Union[str, Dict[str, str]]]:
+        r"""Converts the chunk to the OpenAI format."""
+        if isinstance(self.audio_url, AudioURL):
+            return self.model_dump()
+        else:
+            return {"type": self.type, "audio_url": {"url": self.audio_url}}
```
patrickvonplaten, line 271: nice!

src/mistral_common/protocol/instruct/messages.py
```diff
@@ -171,18 +174,52 @@ def from_audio(cls, audio: Audio) -> "RawAudio":
...
+class AudioURLChunk(BaseContentChunk):
+    r"""Audio URL chunk.
+
+    Attributes:
+        type: The type of the chunk, which is always `ChunkTypes.audio_url`.
+        audio_url: The URL of the audio file.
+    """
+
+    type: Literal[ChunkTypes.audio_url] = ChunkTypes.audio_url
+    audio_url: Union[str, AudioURL]
```
patrickvonplaten, line 226: do we have to allow both cases here?

src/mistral_common/tokens/tokenizers/audio.py
```diff
@@ -178,16 +190,51 @@ def _encode_audio_chunk(self, content: AudioChunk) -> AudioEncoding:
...
         )
 
-    def __call__(self, content: AudioChunk) -> AudioEncoding:
-        r"""Call the encoder on an audio chunk.
+    def _encode_audio_chunk(self, content: AudioChunk) -> AudioEncoding:
+        audio = Audio.from_raw_audio(content.input_audio)
+        return self._encode_audio(audio)
+
+    def _encode_audio_url_chunk(self, content: AudioURLChunk) -> AudioEncoding:
+        url = content.audio_url if isinstance(content.audio_url, str) else content.audio_url.url
```
patrickvonplaten: why do we need to allow both cases (`url` and `AudioURL` object) if we always only need the url as a string?

src/mistral_common/protocol/instruct/messages.py
patrickvonplaten: To begin with can we maybe **only** support `str`?

src/mistral_common/tokens/tokenizers/audio.py
```diff
@@ -178,16 +190,51 @@ def _encode_audio_chunk(self, content: AudioChunk) -> AudioEncoding:
...
+        else:
+            try:
+                audio = Audio.from_base64(url)
+            except Exception:
+                raise ValueError(
+                    (
+                        "Failed to decode audio from the provided AudioURLChunk. Please ensure the URL is a local file "
+                        "path, a valid URL, or a base64-encoded audio string."
+                    )
+                )
```
patrickvonplaten: Ok for now for me!

IMO it's a bit cleaner if we could try to classify the url into one of the different types:

url_type: URLType = retrieve_url_type(...) # returns an enum

and then we do

```py
if audio_type == AudioType.URL:

elif == FILE

elif Base64

....
```

patrickvonplaten (approved):

src/mistral_common/protocol/instruct/messages.py
juliendenize: Think we do, vLLM does second format. I added a `url`property to easily fetch it instead of having lots of if else to get it.

tests/test_converters.py
```diff
@@ -190,6 +202,56 @@ def test_convert_image_url_chunk(openai_image_url_chunk: Dict, image_url_chunk:
...
 
 
+@pytest.mark.parametrize(
+    ["vllm_audio_url_chunk", "audio_url_chunk"],
+    [
+        (
+            {
+                "type": "audio_url",
+                "audio_url": {
+                    "url": "https://freetestdata.com/wp-content/uploads/2021/09/Free_Test_Data_100KB_MP3.mp3"
```
juliendenize:
```suggestion
                    "url": AUDIO_SAMPLE_URL
```

PR #122, opened 2025-07-21
Add think chunk

juliendenize:
The main goal of this PR is to add the `ThinkChunk`.

It required adding also:
- support of list of contents in system and assistant messages
- support of TextChunk and ThinkChunk for system and assistant messages

-------

`to_openai` and `from_openai` are losing their meaning since audio_url and now think chunks as openai do not support those.


src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -1010,6 +1044,75 @@ def _encode_tool_calls_in_assistant_message(self, message: AssistantMessageType)
...
+
+        # Encode content
+        curr_tokens = []
+        if isinstance(message.content, str):
+            curr_tokens += self.tokenizer.encode(message.content.rstrip(" "), bos=False, eos=False)
+        elif isinstance(message.content, list):
+            for chunk in message.content:
+                if isinstance(chunk, TextChunk):
+                    curr_tokens += self.tokenizer.encode(chunk.text.rstrip(" "), bos=False, eos=False)
+
```
juliendenize:
```suggestion
```

src/mistral_common/protocol/instruct/normalize.py
```diff
@@ -146,9 +147,24 @@ def _aggregate_assistant_messages(self, messages: List[UATS]) -> AssistantMessag
                     tool_calls.append(normalized_tool_call)
 
             if message.content:
-                aggregated_content.append(self._aggregate_content_chunks(message.content))
-
+                if isinstance(message.content, str):
```
patrickvonplaten: Uff I don't love this. We should make use of `_aggregate_content_chunks` here? Why not use this function?

src/mistral_common/protocol/instruct/normalize.py
```diff
@@ -130,7 +131,7 @@ def _normalize_tool_call(self, tool_call: ToolCall) -> ToolCall:
         )
 
     def _aggregate_assistant_messages(self, messages: List[UATS]) -> AssistantMessageType:
-        aggregated_content: List[str] = []
+        aggregated_content: List[Union[TextChunk, ThinkChunk]] = []
```
patrickvonplaten: Think we should still allow `str` no?

src/mistral_common/protocol/instruct/normalize.py
```diff
@@ -370,6 +398,36 @@ def normalizer() -> "InstructRequestNormalizerV13":
...
+    ) -> Sequence[UATS]:
+        if role == Roles.system:
+            return [
+                self._system_message_class(content=self._aggregate_content_chunks(message.content))
+                for message in messages
+                if isinstance(message, self._system_message_class)
+            ]
+        return super()._aggregate_role(messages, role, latest_call_ids)
+
+    def _aggregate_content_chunks(  # type: ignore[override]
```
patrickvonplaten:
```suggestion
    def _aggregate_content_chunks(
```

no type ignore please

src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -995,6 +1015,20 @@ class InstructTokenizerV13(InstructTokenizerV11):
...
     _user_message_position_to_encode_tools = UserMessagePosition.first
 
+    def __init__(
+        self,
+        tokenizer: Tokenizer,
+        image_encoder: Optional[ImageEncoder] = None,
+        audio_encoder: Optional[AudioEncoder] = None,
+    ) -> None:
+        super().__init__(tokenizer, image_encoder, audio_encoder)
+        try:
```
patrickvonplaten: no try ... except here please. Better to instead make use of private attribute: `tokenizer._all_special_tokens`

src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -1010,6 +1044,74 @@ def _encode_tool_calls_in_assistant_message(self, message: AssistantMessageType)
...
+                tokens += self.tokenizer.encode(chunk.text, bos=False, eos=False)
+            elif isinstance(chunk, ThinkChunk):
+                tokens += self.encode_think(chunk)
+            else:
+                raise ValueError(f"Unsupported chunk type: {type(chunk)}")
+
+        tokens.append(self.END_SYSTEM)
+        return tokens
+
+    def encode_assistant_message(
```
patrickvonplaten: too much DUP code. Let's adapt previous encode_assistante_message function so that we only have to add an `encode_think(...)` to the V13 tokenizer

src/mistral_common/protocol/instruct/normalize.py
juliendenize: It's still supported changed the name of variables to better reflect that

src/mistral_common/protocol/instruct/normalize.py
```diff
@@ -318,7 +358,11 @@ def _aggregate_role(
         elif role == Roles.user:
             return [self._aggregate_user_messages(messages)]
         elif role == Roles.system:
-            return messages
+            return [
+                SystemMessage(content=self._aggregate_content_chunks(message.content))
+                for message in messages
+                if isinstance(message, self._system_message_class)
```
patrickvonplaten:
```suggestion
                SystemMessage(content=self._aggregate_content_chunks(message.content))
                for message in messages
                if isinstance(message, self._system_message_class)
```

can we do a `self._aggregate_system_messages(...)` here?

patrickvonplaten: Also there might be multiple system messages I think

src/mistral_common/protocol/instruct/normalize.py
```diff
@@ -370,6 +414,17 @@ def normalizer() -> "InstructRequestNormalizerV13":
...
+    def _aggregate_role(
+        self, messages: List[UATS], role: Optional[Roles], latest_call_ids: List[str]
+    ) -> Sequence[UATS]:
+        if role == Roles.system:
+            return [
+                self._system_message_class(content=self._aggregate_content_chunks(message.content))
+                for message in messages
+                if isinstance(message, self._system_message_class)
+            ]
+        return super()._aggregate_role(messages, role, latest_call_ids)
```
patrickvonplaten: Hmm I'd just add a self.aggregate_system_messages here and also one before (that raises if there are chunks in the system message

src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -794,11 +815,20 @@ def encode_system_message(self, message: SystemMessage) -> List[int]:
...
+        if isinstance(message.content, str):
+            tokens += self.tokenizer.encode(message.content, bos=False, eos=False)
+        else:
+            for chunk in message.content:
+                if isinstance(chunk, TextChunk):
+                    tokens += self.tokenizer.encode(chunk.text, bos=False, eos=False)
+                elif isinstance(chunk, ThinkChunk):
+                    tokens += self.encode_think(chunk)
+                else:
+                    raise ValueError(f"Unsupported chunk type: {type(chunk)}")
```
patrickvonplaten:
```suggestion
    tokens = self.encode_content_chunks(...)
```

wouldn't this work?

src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -937,8 +967,17 @@ def encode_assistant_message(
...
+            if isinstance(message.content, str):
+                curr_tokens = self._encode_normal_content_assistant_message(message)
+            elif isinstance(message.content, list):
+                for chunk in message.content:
+                    if isinstance(chunk, TextChunk):
+                        curr_tokens += self.tokenizer.encode(chunk.text.rstrip(" "), bos=False, eos=False)
+
+                    elif isinstance(chunk, ThinkChunk):
+                        curr_tokens += self.encode_think(chunk)
+                    else:
```
patrickvonplaten: Why not use `encode_content_chunks(...)`?

src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -1028,3 +1085,19 @@ def encode_tool_message(self, message: ToolMessage, is_before_last_user_message:
             self.END_TOOL_RESULTS,
         ]
         return curr_tokens
+
+    def encode_think(self, chunk: ThinkChunk) -> List[int]:
```
patrickvonplaten, line 1141: nice!

src/mistral_common/protocol/instruct/normalize.py
```diff
@@ -90,18 +91,48 @@ def _normalize_json_content(self, content: Optional[str]) -> str:
             normalized_content = content
         return normalized_content
 
-    def _aggregate_content_chunks(self, content: Union[str, List[TextChunk]], chunk_join_str: str = "\n\n") -> str:
+    @overload
+    def _aggregate_content_chunks(
+        self, content: List[Union[str, TextChunk, ThinkChunk]], chunk_join_str: str = "\n\n"
```
patrickvonplaten:
```suggestion
        self, content: List[Union[str, TextChunk, ThinkChunk]]) -> ...
        chunk_join_str = CHUNK_JOIN_STR
```

`chunk_join_str` seems to be always the same. Can we maybe it a constant variable instead of allowing the user to pass it?

src/mistral_common/protocol/instruct/normalize.py
```diff
@@ -90,18 +91,48 @@ def _normalize_json_content(self, content: Optional[str]) -> str:
...
+            for chunk in content:
+                if isinstance(chunk, str):
+                    chunk = TextChunk(text=chunk)
+
+                if isinstance(chunk, TextChunk):
+                    if aggregated_content and isinstance(aggregated_content[-1], TextChunk):
+                        aggregated_content[-1].text += chunk_join_str + chunk.text
+                    else:
+                        aggregated_content.append(chunk)
+                elif isinstance(chunk, ThinkChunk):
```
patrickvonplaten: what if two think chunks follow each other - is this allowed? Should we maybe throw in this case?

src/mistral_common/protocol/instruct/normalize.py
```diff
@@ -90,18 +91,48 @@ def _normalize_json_content(self, content: Optional[str]) -> str:
...
+                    aggregated_content.append(chunk)
+                else:
+                    raise ValueError(f"Unsupported chunk type {type(chunk)}")
+
+            if len(aggregated_content) == 1 and isinstance(aggregated_content[0], TextChunk):
+                return aggregated_content[0].text
```
patrickvonplaten:
```suggestion
            if len(aggregated_content) == 1 and isinstance(aggregated_content[0], TextChunk):
                return aggregated_content[0].text
```

is it necessary to have this statement here? Couldn't we just pass the single TextChunk in this case?

src/mistral_common/protocol/instruct/normalize.py
```diff
@@ -145,10 +177,13 @@ def _aggregate_assistant_messages(self, messages: List[UATS]) -> AssistantMessag
                     normalized_tool_call = self._normalize_tool_call(tool_call)
                     tool_calls.append(normalized_tool_call)
 
-            if message.content:
-                aggregated_content.append(self._aggregate_content_chunks(message.content))
+            if isinstance(message.content, str):
+                messages_contents.append(message.content)
+            elif message.content is not None:
+                messages_contents.extend(message.content)
```
patrickvonplaten:
```suggestion
            if (content := message.content) is not None:
                messages_contents.extend([content] if isinstance(content, str) else content)
```

patrickvonplaten: first statement is useless no?

src/mistral_common/tokens/tokenizers/instruct.py
patrickvonplaten: It'd be nice to make it work because I think otherwise it's duplicated

src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -646,6 +667,69 @@ def encode_assistant_message(
...
+        self, chunk: Union[str, TextChunk, ThinkChunk], rstrip: bool
+    ) -> Tuple[List[int], None]: ...
+    @overload
+    def _encode_content_chunk(
+        self, chunk: Union[ImageChunk, ImageURLChunk], rstrip: bool
+    ) -> Tuple[List[int], np.ndarray]: ...
+    @overload
+    def _encode_content_chunk(
+        self, chunk: Union[AudioChunk, AudioURLChunk], rstrip: bool
+    ) -> Tuple[List[int], Audio]: ...
```
patrickvonplaten:
```suggestion
    @overload
    def _encode_content_chunk(
        self, chunk: Union[str, TextChunk, ThinkChunk], rstrip: bool
    ) -> Tuple[List[int], None]: ...
    @overload
    def _encode_content_chunk(
        self, chunk: Union[ImageChunk, ImageURLChunk], rstrip: bool
    ) -> Tuple[List[int], np.ndarray]: ...
    @overload
    def _encode_content_chunk(
        self, chunk: Union[AudioChunk, AudioURLChunk], rstrip: bool
    ) -> Tuple[List[int], Audio]: ...
```

bit of a nit, but I'd prefer to just always return `list[int], list[Images], list[Audio]`

patrickvonplaten: But ok, let's leave as is. We can adapt later if needed

src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -646,6 +667,69 @@ def encode_assistant_message(
...
+    def _encode_content_chunks(
+        self, content: Sequence[ContentChunk], allowed_types: Optional[Tuple[Type[ContentChunk], ...]], rstrip: bool
+    ) -> Tuple[List[int], List[np.ndarray], List[Audio]]:
+        tokens: List[int] = []
+        images: List[np.ndarray] = []
+        audio: List[Audio] = []
+
+        for chunk in content:
+            if allowed_types and not isinstance(chunk, allowed_types):
+                raise ValueError(f"Invalid chunk type: {chunk}")
```
patrickvonplaten:
```suggestion
            if allowed_types and not isinstance(chunk, allowed_types):
                raise ValueError(f"Invalid chunk type: {chunk}")
```

ok! Guess we this should normally be in a validator but that works for now!

tests/test_tokenizer_v13.py
```diff
@@ -128,11 +183,140 @@ def test_end_to_end_v13(
...
+            ),
+            "[SYSTEM_PROMPT]S1S3[THINK]TS[/SYSTEM_PROMPT]",
+        ),
+    ],
+)
+def test_encode_system_message(
+    v13_tekkenizer_think: InstructTokenizerV13, message: SystemMessage, expected: str
+) -> None:
+    encoded = v13_tekkenizer_think.encode_system_message(message)
+    assert v13_tekkenizer_think.decode(encoded, special_token_policy=SpecialTokenPolicy.KEEP) == expected
```
patrickvonplaten, line 322: sorry can we add one more end-to-end test that does both normalization and encoding (for more complex system messages)

src/mistral_common/protocol/instruct/normalize.py
juliendenize: It avoids backward incomp

src/mistral_common/tokens/tokenizers/instruct.py
juliendenize: This also helps mypy.

src/mistral_common/protocol/instruct/normalize.py
juliendenize: Maybe, let's add this in a subsequent PR i'm adding a todo

src/mistral_common/tokens/tokenizers/instruct.py
patrickvonplaten: Hmm yeah don't love it - makes the code hard to read IMO, but it's all private methods so could be changed after

patrickvonplaten: Mypy would also work with a single encode_content_chunks I think

src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -794,13 +867,48 @@ def encode_system_message(self, message: SystemMessage) -> List[int]:
...
-            self.END_SYSTEM,
-        ]
+
+        tokens = [self.BEGIN_SYSTEM]
+        if isinstance(message.content, str):
+            tokens += self._encode_content_chunk(message.content, rstrip=True)[0]
+        else:
+            tokens += self._encode_content_chunks(message.content, allowed_types=(TextChunk, ThinkChunk), rstrip=True)[
+                0
+            ]
```
patrickvonplaten:
```suggestion
        if isinstance(content := message.content, str):
            content = [TextChunk(text=content)]

        tokens += self._encode_content_chunks(content, allowed_types=(TextChunk, ThinkChunk), rstrip=True)[0]
```

would be a tad cleaner but it's a nit

src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -937,8 +1046,12 @@ def encode_assistant_message(
...
         curr_tokens: list = []
         if message.content:
-            assert isinstance(message.content, str), f"Message content must be a string. Got {message.content}"
-            curr_tokens += self._encode_normal_content_assistant_message(message)
+            if isinstance(message.content, str):
+                curr_tokens = self._encode_normal_content_assistant_message(message)
+            elif isinstance(message.content, list):
+                curr_tokens += self._encode_content_chunks(
+                    message.content, allowed_types=(TextChunk, ThinkChunk), rstrip=True
+                )[0]
```
patrickvonplaten: Think this could also be simplified by normalizing everyhing to list[Chunk]

src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -1028,3 +1159,19 @@ def encode_tool_message(self, message: ToolMessage, is_before_last_user_message:
...
+        r"""Encode a thinking chunk.
+
+        Args:
+            chunk: The thinking chunk to encode.
+        Returns:
+            The encoded tokens.
+        """
+        assert self.BEGIN_THINK is not None, "think tokens are not available for this tokenizer."
+        assert self.END_THINK is not None, "think tokens are not available for this tokenizer."
+        tokens = self.tokenizer.encode(chunk.thinking.rstrip(" "), bos=False, eos=False)
```
joa-stdn: I think we should remove rstrip no?

src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -646,6 +667,69 @@ def encode_assistant_message(
...
+    ) -> Tuple[List[int], np.ndarray, None]: ...
+    @overload
+    def _encode_content_chunk(
+        self, chunk: Union[AudioChunk, AudioURLChunk], rstrip: bool
+    ) -> Tuple[List[int], None, Audio]: ...
+    def _encode_content_chunk(
+        self, chunk: Union[str, ContentChunk], rstrip: bool
+    ) -> Union[Tuple[List[int], Optional[np.ndarray], Optional[Audio]]]:
+        if isinstance(chunk, str):
+            text = chunk.rstrip(" ") if rstrip else chunk
```
joa-stdn: Only `rstrip` for str for assistant messages to be consistent maybe?

src/mistral_common/protocol/instruct/normalize.py
```diff
@@ -90,18 +93,49 @@ def _normalize_json_content(self, content: Optional[str]) -> str:
...
         if isinstance(content, list):
-            return chunk_join_str.join([chunk.text for chunk in content])
-        else:
+            aggregated_content: List[Union[TextChunk, ThinkChunk]] = []
+            for chunk in content:
+                if isinstance(chunk, str):
+                    chunk = TextChunk(text=chunk)
+
+                if isinstance(chunk, TextChunk):
+                    # TODO(Julien): Add a check for previous text chunks especially if one is open
```
joa-stdn: I think you should add this to `validator.py`, not here no?

src/mistral_common/protocol/instruct/normalize.py
```diff
@@ -90,18 +93,49 @@ def _normalize_json_content(self, content: Optional[str]) -> str:
...
+                        aggregated_content.append(chunk)
+                elif isinstance(chunk, ThinkChunk):
+                    aggregated_content.append(chunk)
+                else:
+                    raise ValueError(f"Unsupported chunk type {type(chunk)}")
+
+            if len(aggregated_content) == 1 and isinstance(aggregated_content[0], TextChunk):
+                return aggregated_content[0].text
+            return aggregated_content
+        elif isinstance(content, str):
```
joa-stdn: (nit) maybe move this up as first condition in if as this is one line?

src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -794,13 +867,48 @@ def encode_system_message(self, message: SystemMessage) -> List[int]:
...
+
+        if isinstance(content, str):
+            return super().encode_user_content(content, is_last, system_prompt)
+
+        has_one_img_one_text_first = len(content) == 2 and isinstance(content[1], (ImageChunk, ImageURLChunk))
+        if force_img_first and has_one_img_one_text_first:
+            # make sure that if exactly one image and text chunk are passed we force the image chunk to be first
+            content = [content[1], content[0]]
+
+        tokens, images, audio = self._encode_content_chunks(content, allowed_types=None, rstrip=False)
```
joa-stdn: User messages cannot contain `ThinkChunk`, should we validate this somewhere? Maybe pydantic validator in `UserMessage`?

juliendenize: Pydantic should handle this now nice catch !

src/mistral_common/protocol/instruct/normalize.py
juliendenize: Yes the TODO is a reminder i added where to put it.

src/mistral_common/protocol/instruct/normalize.py
juliendenize: nice

src/mistral_common/tokens/tokenizers/instruct.py
juliendenize: not anymore now that I know that rstrip is only for str assistant message but i made it a bit cleaner

tests/test_tokenizer_v13.py
juliendenize: as discussed, i think it's done

src/mistral_common/tokens/tokenizers/instruct.py
juliendenize: Turns out it was highly unnecessary with all the changes done

tests/test_converters.py
```diff
@@ -365,6 +374,23 @@ def test_convert_tool_call() -> None:
             ),
         ),
         (OpenAIAssistantMessage(role="assistant", content="Hi"), AssistantMessage(content="Hi")),
+        (
+            {
+                "role": "assistant",
+                "content": [
+                    {"type": "text", "text": "Hi"},
+                    {"type": "thinking", "thinking": "Hello", "closed": False},
```
joa-stdn: (nit) in the future I think we may want to prohibit such messages because there's a non-closed in the middle
Maybe we reverse 2nd and 3rd?

Issue #125, opened 2025-07-29
[BUG: Can't install dependency]


jacereda:
### Python -VV

```shell
Can't setup the venv.
```

### Pip Freeze

```shell
Can't setup the venv.
```

### Reproduction Steps

Make sure there's no `cmake` package installed.

On a fresh checkout, just try to invoke `uv sync`:

```E:\mistral-common>uv sync
Resolved 116 packages in 1ms
  x Failed to build `sentencepiece==0.2.0`
  |-> The build backend returned an error
  `-> Call to `setuptools.build_meta:__legacy__.build_wheel` failed (exit code: 1)

      [stderr]
      Traceback (most recent call last):
        File "<string>", line 14, in <module>
          requires = get_requires_for_build({})
        File "C:\Users\Jorge\AppData\Local\uv\cache\builds-v0\.tmpgsRVpx\Lib\site-packages\setuptools\build_meta.py",
      line 331, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=[])
                 ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "C:\Users\Jorge\AppData\Local\uv\cache\builds-v0\.tmpgsRVpx\Lib\site-packages\setuptools\build_meta.py",
      line 301, in _get_build_requires
          self.run_setup()
          ~~~~~~~~~~~~~~^^
        File "C:\Users\Jorge\AppData\Local\uv\cache\builds-v0\.tmpgsRVpx\Lib\site-packages\setuptools\build_meta.py",
      line 512, in run_setup
          super().run_setup(setup_script=setup_script)
          ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "C:\Users\Jorge\AppData\Local\uv\cache\builds-v0\.tmpgsRVpx\Lib\site-packages\setuptools\build_meta.py",
      line 317, in run_setup
          exec(code, locals())
          ~~~~^^^^^^^^^^^^^^^^
        File "<string>", line 128, in <module>
        File "C:\Users\Jorge\AppData\Roaming\uv\python\cpython-3.13.5-windows-x86_64-none\Lib\subprocess.py", line
      414, in check_call
          retcode = call(*popenargs, **kwargs)
[...]

Issue #75, opened 2025-01-28
[BUG: Dependency incompatibility with Transformers


odusseys:
### Python -VV

```shell
Python 3.10.16 | packaged by conda-forge | (main, Dec  5 2024, 14:16:10) [GCC 13.3.0]
```

### Pip Freeze

```shell
absl-py @ file:///home/conda/feedstock_root/build_artifacts/absl-py_1733730548347/work
accelerate @ file:///home/conda/feedstock_root/build_artifacts/accelerate_1734379448962/work
addict @ file:///home/conda/feedstock_root/build_artifacts/addict_1636818143388/work
aiofiles @ file:///home/conda/feedstock_root/build_artifacts/aiofiles_1733829030387/work
aiohappyeyeballs @ file:///home/conda/feedstock_root/build_artifacts/aiohappyeyeballs_1733331917983/work
aiohttp @ file:///home/conda/feedstock_root/build_artifacts/aiohttp_1734596888970/work
aiosignal @ file:///home/conda/feedstock_root/build_artifacts/aiosignal_1734342155601/work
aiostream @ file:///home/conda/feedstock_root/build_artifacts/aiostream_1734876708842/work
AIRStore @ file:///root/anaconda3/conda-bld/airstore_1734726025048/work/AIRStore-1.0.0%2B5b4c9ab269df63f03cccf4444403459c.adda70f37f6259bc134f50b8b125f9cd48b9f145.1734613236-py3.10-none-linux_x86_64.whl#sha256=38776bfd0b72c7ea22a93d1ea11351d39683c86450f0ba580b8cceeff1bb5959
airstore_dataset @ file:///root/anaconda3/conda-bld/airstore_dataset_1732652435625/work
albucore @ file:///home/conda/feedstock_root/build_artifacts/albucore_1727019653300/work
albumentations @ file:///home/conda/feedstock_root/build_artifacts/albumentations_1727025741991/work
alembic @ file:///home/conda/feedstock_root/build_artifacts/alembic_1733728406412/work
altair @ file:///home/conda/feedstock_root/build_artifacts/altair-split_1734244716962/work
annotated-types @ file:///home/conda/feedstock_root/build_artifacts/annotated-types_1733247046149/work
antlr4-python3-runtime @ file:///home/conda/feedstock_root/build_artifacts/antlr-python-runtime-meta_1638309185939/work
anyio @ file:///home/conda/feedstock_root/build_artifacts/anyio_1733532585909/work
[...]

PR #126, opened 2025-07-30
Make sentencepiece optional and refactor optional imports

juliendenize:
This PR makes `sentencepiece` optional to fix issues rising: #125 #75 
This is ok as our recent models are only released as `Tekkenizer`s.

It also refactors optional import to add the functions:
- `in_package_installed`
- `assert_package_installed`

and avoid duplicating code across the codebase.

### DIFF
--- a/src/mistral_common/imports.py
+++ b/src/mistral_common/imports.py
@@ -6,6 +6,10 @@ from typing import Optional
 logger = logging.getLogger(__name__)
 
 
+def _get_dependency_error_message(package_name: str, dependency_group: str) -> str:
+    return f"`{package_name}` is not installed. Please install it with `pip install mistral-common[{dependency_group}]`"
+
+
 def is_package_installed(package_name: str) -> bool:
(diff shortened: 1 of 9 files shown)


src/mistral_common/imports.py
```diff
@@ -0,0 +1,88 @@
...
+
+@lru_cache()
+def is_soxr_installed() -> bool:
+    return is_package_installed("soxr")
+
+
+def assert_hf_hub_installed() -> None:
+    assert_package_installed(
+        "huggingface_hub",
+        "`huggingface_hub` is not installed. Please install it with `pip install mistral-common[hf-hub]`",
```
patrickvonplaten: Maybe we don't have to duplicate the string everywere - one more helper function here?

patrickvonplaten (approved):
Nice!

src/mistral_common/imports.py
juliendenize: nice, let's apply DRY. I added the helper function.

PR #147, opened 2025-10-25
test: add image normalization tests

PrasanaaV:
test: add image normalization tests


juliendenize (changes requested):
Hey thanks for the PR and sorry i took this long to answer, looks good but i think the file where you put the test is not the right one.

test_normalization.py are for normalization of requests and not for image normalization. i'd suggest to move these changes to test_multimodal.py

PR #150, opened 2025-11-10
Add Tokenizer Comparison Script

pandora-s-git:
This script compares the basic `.encode` tokenization between Hugging Face and Mistral Common tokenizers across multiple datasets.


scripts/compare_tokenizer.py
```diff
@@ -0,0 +1,218 @@
...
+' O' -> 44 times
+' Jimmy' -> 32 times
+' (' -> 22 times
+'/' -> 21 times
+'Re' -> 18 times
+'Produ' -> 18 times
+' and' -> 16 times
+'gan' -> 16 times
+'Reg' -> 16 times
+"""
```
juliendenize: Nit: would put that in the docstrings of the function producing the output

scripts/compare_tokenizer.py
```diff
@@ -0,0 +1,218 @@
...
+from collections import Counter
+import json
+
+def compare_tokenizers(
+    hf_tokenizer: AutoTokenizer,
+    mc_tokenizer: MistralTokenizer,
+    content: str,
+) -> Dict[str, Any]:
+    """Compare tokenization between Hugging Face and Mistral tokenizers."""
+    content = content[:10_000]  # Limit to 10000 characters to avoid long sequences
```
juliendenize: Could we actually not hard code this restriction ? Would probably be better to have a default that user can override via cli.

scripts/compare_tokenizer.py
```diff
@@ -0,0 +1,218 @@
...
+    python3 compare_tokenizer.py --hf_model mistralai/Mistral-Nemo-Instruct-2407 --mc_model mistralai/Mistral-Small-3.1-24b-Instruct-2503 --num_samples 1000
+"""
+
+import argparse
+from transformers import AutoTokenizer
+from mistral_common.tokens.tokenizers.mistral import MistralTokenizer
+from mistral_common.tokens.tokenizers.base import SpecialTokenPolicy
+from datasets import load_dataset
+from tqdm import tqdm
+from typing import List, Dict, Any
```
juliendenize: Use modern python annotation so get ridden of List Dict

juliendenize (commented):
Thanks a bunch for these scripts, left few comments. For general feedback and discussions:

- Can you run pre-commit to silence the tests and get rid of linter/formatter issues ?
- As we plan to also host chat templates, wdyt about creating a folder to host scripts, files dedicated to integrations in libraries ?

PR #163, opened 2025-11-27
Add chat templates integration

juliendenize:
This PR aims to add chat templates inside this repo as a valid working implementation in comparison with mistral-common to work with libraries that require jinja templates (llama.cpp, vanilla Transformers, SG Lang ...).

This is in order to give the possibility to our community to use our models beyond the scope of mistral-common when the latter is not integrated in parts of the ecosystem.

We validate these chat templates on a set of dummy requests.

Next step is to add tests for edge cases and errors + integration tests on some datasets.


src/mistral_common/protocol/instruct/messages.py
```diff
@@ -279,6 +279,8 @@ class ToolMessage(BaseMessage):
     def to_openai(self) -> dict[str, str | list[dict[str, str | dict[str, Any]]]]:
         r"""Converts the message to the OpenAI format."""
         assert self.tool_call_id is not None, "tool_call_id must be provided for tool messages."
+        if self.name is not None:
+            return self.model_dump()
```
juliendenize: this is not good should probably revert and add it when need in the tests.

tests/data/chat_templates/v15_image.jinja
```diff
@@ -0,0 +1,204 @@
...
+            {{- available_tools }}
+            {{- model_settings }}
+            {%- set ns.available_tools_and_settings_emitted = true %}
+        {%- endif %}
+        {%- if message['content'] is string %}
+            {{- '[INST]' + message['content'] + '[/INST]' }}
+        {%- elif message['content'] | length > 0 %}
+            {{- '[INST]' }}
+            {%- if message['content'] | length == 2 %}
+                {%- set blocks = message['content'] | sort(attribute='type') %}
```
joa-stdn: is this because of image reordering? written this way is weird no?

tests/data/chat_templates/v15_image.jinja
```diff
@@ -0,0 +1,204 @@
...
+{%- set ns_order = namespace(previous_role=none) %}
+{%- for message in loop_messages %}
+    {%- set current_role = message['role'] %}
+    {%- if ns_order.previous_role is not none %}
+        {%- if ns_order.previous_role == 'system' %}
+            {%- if current_role != 'user' and current_role != 'assistant' and current_role != 'system' %}
+                {{- raise_exception('Unexpected role \'' + current_role + '\' after role \'' + ns_order.previous_role + '\'') }}
+            {%- endif %}
+        {%- elif ns_order.previous_role == 'user' %}
+            {%- if current_role != 'assistant' and current_role != 'system' and current_role != 'user' %}
```
joa-stdn: maybe better to do equality here?

tests/data/chat_templates/v15_image.jinja
```diff
@@ -0,0 +1,204 @@
...
+            {%- for msg in ns_agg.current_group %}
+                {%- if msg['content'] is string %}
+                    {%- set ns_c.text_parts = ns_c.text_parts + [msg['content']] %}
+                {%- elif msg['content'] is not none %}
+                    {%- for block in msg['content'] %}
+                        {%- if block['type'] == 'text' %}
+                            {%- set ns_c.text_parts = ns_c.text_parts + [block['text']] %}
+                        {%- else %}
+                            {%- if ns_c.text_parts | length > 0 %}
+                                {%- set ns_c.chunks = ns_c.chunks + [{'type': 'text', 'text': ns_c.text_parts | join('\n\n')}] %}
```
joa-stdn, line 67: actually internally we only aggregate across messages now but used indeed to aggregate for chunks in the same message

tests/data/chat_templates/v15_image.jinja
```diff
@@ -0,0 +1,204 @@
...
+    {%- elif message['role'] == 'system' %}
+        {{- '[SYSTEM_PROMPT]' -}}
+        {%- if message['content'] is string %}
+            {{- message['content'] -}}
+        {%- else %}
+            {%- for block in message['content'] %}
+                {%- if block['type'] == 'text' %}
+                    {{- block['text'] }}
+                {%- else %}
+                    {{- raise_exception('Only text chunks are supported in system message contents.') }}
```
joa-stdn: do you think we could have a macro for some of those things? like render_content?

tests/data/chat_templates/v15_image_think.jinja
```diff
@@ -0,0 +1,224 @@
...
+
+        {%- if message['content'] is string and message['content'] != '' %}
+            {{- message['content'] }}
+        {%- elif message['content'] | length > 0 %}
+            {%- for block in message['content'] %}
+                {%- if block['type'] == 'text' %}
+                    {{- block['text'] }}
+                {%- elif block['type'] == 'thinking' %}
+                    {{- '[THINK]' + block['thinking'] }}
+                    {%- if block.get('closed', true) %}{{- '[/THINK]' }}{%- endif %}
```
joa-stdn: tbh it should really always be true IMO, are we sure it's going to be ok?

juliendenize: it's an option from mistral-common so i'd like to replicate, think it should be fine anyway as i think very few people would pass this flag.

.github/workflows/lint_build_test.yaml
```diff
@@ -95,3 +95,32 @@ jobs:
...
+      matrix:
+        python-version: [
+          "3.10",
+          "3.11",
+          "3.12",
+          "3.13",
+          "3.14",
+        ]
+    steps:
+      - name: Checkout
```
joa-stdn, line 112: are we ok with github actions?

src/mistral_common/integrations/chat_templates/chat_templates.py
```diff
@@ -0,0 +1,60 @@
...
+from mistral_common.tokens.tokenizers.base import TokenizerVersion
+
+
+def generate_chat_template(
+    spm: bool,
+    tokenizer_version: TokenizerVersion,
+    image_support: bool,
+    audio_support: bool,
+    thinking_support: bool,
+    default_system_prompt: str | None = None,
```
joa-stdn: let's remove defaults?

src/mistral_common/integrations/chat_templates/chat_templates.py
```diff
@@ -0,0 +1,60 @@
...
+        image_support=image_support,
+        audio_support=audio_support,
+        thinking_support=thinking_support,
+        plain_thinking_support=plain_thinking_support,
+        use_token_variables=use_token_variables,
+    )
+    template = _build_chat_template(config)
+
+    if default_system_prompt is not None:
+        escaped_prompt = default_system_prompt.replace("\\", "\\\\").replace("'", "\\'")
```
joa-stdn, line 56: we're not afraid of the thinking chunk here? Worth a comment at least maybe?

src/mistral_common/integrations/chat_templates/template_generator.py
```diff
@@ -0,0 +1,1674 @@
...
+        True
+    """
+
+    version: TokenizerVersion
+    spm: bool = False
+    image_support: bool = False
+    audio_support: bool = False
+    thinking_support: bool = False
+    plain_thinking_support: bool = False
+    use_token_variables: bool = True
```
joa-stdn: this name is very weird, I wonder if we can maybe find a better name no?

src/mistral_common/integrations/chat_templates/template_generator.py
```diff
@@ -0,0 +1,1674 @@
...
+            raise ValueError("Plain thinking support is only available for tokenizer version v11")
+        if self.audio_support and self.plain_thinking_support:
+            raise ValueError("Audio and plain thinking support are mutually exclusive")
+
+    @property
+    def bos_expr(self) -> str:
+        r"""Jinja expression for the BOS token."""
+        if self.use_token_variables:
+            return "bos_token"
+        return "'" + _BOS + "'"
```
joa-stdn: is this better than f strings?

src/mistral_common/integrations/chat_templates/template_generator.py
```diff
@@ -0,0 +1,1674 @@
...
+                "        {{- raise_exception('Only user, assistant, system and tool roles are supported, got ' + message['role'] + '.') }}"  # noqa: E501
+            )
+        else:
+            lines.append(
+                "        {{- raise_exception('Only user, assistant and system roles are supported, got ' + message['role'] + '.') }}"  # noqa: E501
+            )
+    else:
+        # pre-v7: system messages are already filtered out
+        if config.has_tools:
+            if config.version == TokenizerVersion.v2 or (config.spm and config.version == TokenizerVersion.v3):
```
joa-stdn: to me it's quite weird to have both explicit versions and the simplifying properties, we should go one way or the other no?

tests/integrations/chat_templates/transformers/test_core_parity.py
```diff
@@ -0,0 +1,310 @@
...
+            spm=spm, tokenizer_version=version, image_support=image, audio_support=audio, thinking_support=think
+        )
+        if version <= TokenizerVersion.v2:
+            for conv in conversations:
+                for message in conv.messages:
+                    if isinstance(message, (UserMessage, AssistantMessage)) and isinstance(message.content, list):
+                        assert len(message.content) == 1 and isinstance(message.content[0], TextChunk), (
+                            "Only text content is supported for v1 and v2"
+                        )
+                        message.content = str(message.content[0].text)
```
joa-stdn, line 74: is it better to do this or to have them already in str?

tests/integrations/chat_templates/transformers/test_core_parity.py
```diff
@@ -0,0 +1,310 @@
...
+        for conversation in conversations:
+            if version == TokenizerVersion.v2:
+                for message in conversation.messages:
+                    if isinstance(message, ToolMessage):
+                        message.name = "tool"
```
joa-stdn: I think we can just keep names all the time?

tests/integrations/chat_templates/transformers/test_core_parity.py
```diff
@@ -0,0 +1,310 @@
...
+        # Consecutive user messages should be aggregated (not raise an error)
+        valid_consecutive_users = {
+            "messages": [
+                {"role": "user", "content": "Hello"},
+                {"role": "user", "content": "World"},
+                {"role": "assistant", "content": "Hi"},
+            ]
+        }
+        # This should not raise
+        encode_transformers(chat_template, valid_consecutive_users)
```
joa-stdn: why don't we check matching here?

tests/integrations/chat_templates/transformers/test_core_parity.py
```diff
@@ -0,0 +1,310 @@
...
+    def test_invalid_chunks(
+        self,
+        spm: bool,
+        version: TokenizerVersion,
+        image: bool,
+        audio: bool,
+        think: bool,
+    ) -> None:
+        invalid_sp_think = {
+            "messages": [
```
joa-stdn: maybe let's define this in fixtures / outside of tests and have the test itself shorter?

tests/integrations/chat_templates/transformers/test_core_parity.py
```diff
@@ -0,0 +1,310 @@
...
+        spm: bool,
+        version: TokenizerVersion,
+        mode: ValidationMode,
+        image: bool,
+        audio: bool,
+        think: bool,
+    ) -> None:
+        conversations = _get_conversations(version, mode, image, audio, think)
+
+        mistral_tokenizer = _get_mistral_tokenizer(
```
joa-stdn: curious why we don't do the MistralTokenizer.from_file directly with a file in tmp_path etc?

tests/integrations/chat_templates/conftest.py
```diff
@@ -0,0 +1,86 @@
...
+from unittest.mock import patch
+
+import pytest
+
+from mistral_common.tokens.tokenizers.base import TokenizerVersion
+from tests.integrations.chat_templates.helpers import _IMAGE
+
+# All configurations for output comparison tests (including SPM).
+# Tuples: (version, spm, image, audio, think, plain_think)
+ALL_CONFIGS: list[tuple[TokenizerVersion, bool, bool, bool, bool, bool]] = [
```
joa-stdn: should we dataclass for this?

tests/integrations/chat_templates/helpers.py
```diff
@@ -0,0 +1,340 @@
...
+        render_kwargs["reasoning_effort"] = reasoning_effort
+
+    return jinja_template.render(**render_kwargs)
+
+
+def encode_mistral_common(mistral_tokenizer: MistralTokenizer, chat_request: ChatCompletionRequest, spm: bool) -> str:
+    r"""Encode a chat request using mistral-common tokenizer."""
+    mistral_encoded = str(mistral_tokenizer.encode_chat_completion(chat_request).text)
+    # Remove image tokens except one per image
+    mistral_encoded = mistral_encoded.replace("[IMG]", "").replace("[IMG_BREAK]", "").replace("[IMG_END]", "[IMG]")
```
joa-stdn, line 127: Actually doesn't it make more sense to concatenate all [IMG] into [IMG] and drop the two others? Maybe we don't care, this is nit

tests/integrations/chat_templates/helpers.py
```diff
@@ -0,0 +1,340 @@
...
+    # Only add reasoning_effort for v15+ templates that support it
+    if reasoning_effort is not None or "reasoning_effort" in template:
+        render_kwargs["reasoning_effort"] = reasoning_effort
+
+    return jinja_template.render(**render_kwargs)
+
+
+def encode_mistral_common(mistral_tokenizer: MistralTokenizer, chat_request: ChatCompletionRequest, spm: bool) -> str:
+    r"""Encode a chat request using mistral-common tokenizer."""
+    mistral_encoded = str(mistral_tokenizer.encode_chat_completion(chat_request).text)
```
joa-stdn, line 123: there's no way to compare token ids? I thought this was the purpose...

tests/integrations/chat_templates/helpers.py
```diff
@@ -0,0 +1,340 @@
...
+
+    render_kwargs: dict[str, Any] = {
+        "messages": messages,
+        "tools": tools,
+        "bos_token": "<s>",
+        "eos_token": "</s>",
+    }
+
+    # Only add reasoning_effort for v15+ templates that support it
+    if reasoning_effort is not None or "reasoning_effort" in template:
```
joa-stdn: seems weird to me, I would have said that we need to restructure the tests so that this line can just be
```
if reasoning_effort is not None
```
no?

tests/integrations/chat_templates/helpers.py
```diff
@@ -0,0 +1,340 @@
...
+    mistral_encoded = mistral_encoded.replace("[IMG]", "").replace("[IMG_BREAK]", "").replace("[IMG_END]", "[IMG]")
+    # Remove audio tokens except one per audio
+    mistral_encoded = mistral_encoded.replace("[AUDIO]", "").replace("[BEGIN_AUDIO]", "[AUDIO]")
+    if spm:
+        mistral_encoded = mistral_encoded.replace(SPM_WHITESPACE, " ").replace("<0x0A>", "\n")
+    return mistral_encoded
+
+
+def encode_transformers(
+    chat_template: str, chat_request: ChatCompletionRequest | dict[str, Any], keep_name_for_tools: bool = False
```
joa-stdn: why do we allow both pydantic object and ChatCompletionRequest

tests/integrations/chat_templates/helpers.py
```diff
@@ -0,0 +1,340 @@
...
+        openai_request = chat_request
+    for tool in openai_request.get("tools", []):
+        tool["function"].pop("strict", False)
+
+    reasoning_effort = openai_request.get("reasoning_effort")
+    template_kwargs: dict[str, Any] = {}
+    if reasoning_effort is not None:
+        template_kwargs["reasoning_effort"] = reasoning_effort
+
+    encoded = render_jinja_template(
```
joa-stdn: maybe add a comment on why we have render_template as well?

tests/integrations/chat_templates/helpers.py
```diff
@@ -0,0 +1,340 @@
...
+            return InstructTokenizerV11
+        case TokenizerVersion.v13:
+            return InstructTokenizerV13
+        case TokenizerVersion.v15:
+            return InstructTokenizerV15
+        case _:
+            raise ValueError(f"Unknown tokenizer version: {tokenizer_version}")
+
+
+def _get_mistral_tekkenizer(
```
joa-stdn: why don't we use MistralTokenizer.from_file?

tests/integrations/chat_templates/helpers.py
```diff
@@ -0,0 +1,340 @@
...
+    tokenizer_version: TokenizerVersion, validation_mode: ValidationMode, image: bool, audio: bool, think: bool
+) -> MistralTokenizer:
+    r"""Build a `MistralTokenizer` with Tekken backend."""
+    special_tokens = get_special_tokens(tokenizer_version=tokenizer_version, add_audio=audio, add_think=think)
+    with open(MistralTokenizer._data_path() / "tekken_240911.json", "r", encoding="utf-8") as f:
+        json_tekkenizer = json.load(f)
+    vocab = json_tekkenizer["vocab"]
+    vocab_size = json_tekkenizer["config"]["default_vocab_size"]
+    pattern = json_tekkenizer["config"]["pattern"]
+    model_settings_builder = (
```
joa-stdn: this is too hacky no?

tests/data/chat_templates/v15_image.jinja
juliendenize: should be something that i change in the end but should be first deprecated

.github/workflows/lint_build_test.yaml
juliendenize: i should move to bk probably eventually. will adress in followup

src/mistral_common/integrations/chat_templates/chat_templates.py
juliendenize: wdym ? think chunks doesn't need escape right ?
Or do you mean that users should keep in mind to add think tokens if needed ?

src/mistral_common/integrations/chat_templates/template_generator.py
juliendenize: changed the name lmk if it looks better :)

src/mistral_common/integrations/chat_templates/template_generator.py
juliendenize: nop nice catch

src/mistral_common/integrations/chat_templates/template_generator.py
juliendenize: true, added more properties

tests/integrations/chat_templates/transformers/test_core_parity.py
juliendenize: true

tests/integrations/chat_templates/transformers/test_core_parity.py
juliendenize: these tests are redundant because already tested valid ones in the conversations, going to remove.

tests/integrations/chat_templates/transformers/test_core_parity.py
juliendenize: we don't have all the versions in the fixtures data. I could download some official but decided instead to build them on runtime. Is it a blocker to you ?

tests/integrations/chat_templates/helpers.py
juliendenize: it is possible but requires to use apply_chat_template from HF. I could add this indeed.

tests/integrations/chat_templates/helpers.py
juliendenize: the other one was to test invalid or dict request from openai, created another function

tests/integrations/chat_templates/helpers.py
juliendenize: why do you think that ?

tests/integrations/chat_templates/transformers/test_core_parity.py
juliendenize: in the end i changed it to be able to use MistralCommonBackend :)

tests/integrations/chat_templates/helpers.py
juliendenize: added

tests/integrations/chat_templates/transformers/test_core_parity.py
juliendenize: better like this as i also want to test chunks

tests/integrations/chat_templates/helpers.py
juliendenize: harder to setup if you have multiple images that are one after the other

tests/integrations/chat_templates/helpers.py
juliendenize: there is a docstring for the function that should explain it

src/mistral_common/integrations/chat_templates/chat_templates.py
joa-stdn: I meant system prompt themselves can contain chunks? But here it's str, should be ok

src/mistral_common/integrations/chat_templates/template_generator.py
```diff
@@ -0,0 +1,1698 @@
...
+    ]
+
+
+def _generate_message_aggregation(config: TemplateConfig) -> str:
+    r"""Generate message aggregation pre-processing block.
+
+    Aggregates consecutive messages with the same role before the alternation
+    check to match the behavior of the mistral-common normalizer:
+
+    - **User messages**: consecutive messages merged into one, text content joined
```
joa-stdn, line 685: we need to nuke this actually, any chance we can do that in follow up?

src/mistral_common/integrations/chat_templates/template_generator.py
```diff
@@ -0,0 +1,1698 @@
...
+      with `"\\n\\n"`, non-text chunks (image, audio) preserved in order.
+    - **Assistant messages**: consecutive messages merged into one, text content joined
+      with `"\\n\\n"`, `tool_calls` lists concatenated.
```
joa-stdn, line 688: all of this is no longer up to date in new models I think

tests/integrations/chat_templates/helpers.py
```diff
@@ -0,0 +1,695 @@
...
+    audio_encoder = _get_audio_encoder() if audio else None
+    image_encoder = _get_image_encoder(tokenizer) if image else None
+
+    instruct_tokenizer: InstructTokenizer = _get_instruct_tokenizer_class(tokenizer_version)(
+        tokenizer=tokenizer,
+        image_encoder=image_encoder,
+        audio_encoder=audio_encoder,
+    )
+    model_settings_builder = tokenizer.model_settings_builder if isinstance(tokenizer, Tekkenizer) else None
+    return MistralTokenizer(
```
joa-stdn: I think MistralTokenizer.from_file is more natural, we have that in test_tokenizer_v14 etc no?

tests/integrations/chat_templates/helpers.py
```diff
@@ -0,0 +1,695 @@
...
+
+    audio_encoder = _get_audio_encoder() if audio else None
+    image_encoder = _get_image_encoder(tokenizer) if image else None
+
+    instruct_tokenizer: InstructTokenizer = _get_instruct_tokenizer_class(tokenizer_version)(
+        tokenizer=tokenizer,
+        image_encoder=image_encoder,
+        audio_encoder=audio_encoder,
+    )
+    model_settings_builder = tokenizer.model_settings_builder if isinstance(tokenizer, Tekkenizer) else None
```
joa-stdn: this would allow not doing this

tests/integrations/chat_templates/helpers.py
```diff
@@ -0,0 +1,695 @@
...
+
+    for attr_name in vars(instruct_tokenizer):
+        val = getattr(instruct_tokenizer, attr_name)
+        if isinstance(val, int) and 0 <= val < num_special:
+            active_ids.add(val)
+
+    return active_ids
+
+
+def _build_hf_tokenizer(
```
joa-stdn: can do we a file for hf_utils maybe? This one is quite big?

tests/integrations/chat_templates/helpers.py
```diff
@@ -0,0 +1,695 @@
...
+    """
+    tokenizer = instruct_tokenizer.tokenizer
+    # BOS and EOS are always active
+    active_ids: set[int] = {tokenizer.bos_id, tokenizer.eos_id}
+
+    num_special = getattr(tokenizer, "num_special_tokens", 0)
+    if num_special == 0:
+        return active_ids
+
+    for attr_name in vars(instruct_tokenizer):
```
joa-stdn: not super clean no? why don't we just add all special tokens to the HF model? this is what we do anyways in Tekkenizer?

tests/integrations/chat_templates/helpers.py
```diff
@@ -0,0 +1,695 @@
...
+    r"""Encode a chat request using the transformers render_jinja_template.
+
+    Converts a ChatCompletionRequest to OpenAI format and renders through
+    the HuggingFace transformers rendering pipeline.
+    """
+    assert _HAS_TRANSFORMERS, "transformers is required"
+    openai_request = chat_request.to_openai()
+    if keep_name_for_tools:
+        for openai_message, chat_message in zip(openai_request["messages"], chat_request.messages):
+            if chat_message.role == "tool":
```
joa-stdn: this is duplicated can we factorize?

tests/integrations/chat_templates/helpers.py
```diff
@@ -0,0 +1,384 @@
...
+        tokenizer_version,
+    )
+
+    image_encoder = _get_image_encoder(tokenizer) if image else None
+
+    instruct_tokenizer: InstructTokenizer = _get_instruct_tokenizer_class(tokenizer_version)(
+        tokenizer=tokenizer,
+        image_encoder=image_encoder,
+    )
+    return MistralTokenizer(
```
joa-stdn: why not from file here?

joa-stdn (approved):

tests/integrations/chat_templates/helpers.py
juliendenize: Refactored — now uses `MistralTokenizer.from_file` via `_build_tekken_json` + `_get_mistral_tekkenizer`. No more manual tokenizer construction.

PR #164, opened 2025-12-10
Rename get control -> to get special & add is_special

patrickvonplaten:
get_control was bad naming. We want to use `is_special` for vLLM for example here: https://github.com/vllm-project/vllm/blob/bf6a3d0ff5a69e0a30567f2ad417530c002eaa4e/vllm/transformers_utils/tokenizers/mistral.py#L158

### DIFF
--- a/src/mistral_common/tokens/tokenizers/sentencepiece.py
+++ b/src/mistral_common/tokens/tokenizers/sentencepiece.py
@@ -18,7 +18,7 @@ from mistral_common.tokens.tokenizers.image import ImageConfig, MultiModalVersio
 warnings.filterwarnings(
     action="once",
     category=FutureWarning,
-    message=r"*`get_control_token` is deprecated.*",
+    message=r".*`get_control_token` is deprecated.*",
 )
 
 
--- a/src/mistral_common/tokens/tokenizers/tekken.py
+++ b/src/mistral_common/tokens/tokenizers/tekken.py
@@ -22,7 +22,7 @@ from mistral_common.tokens.tokenizers.image import ImageConfig
 warnings.filterwarnings(
     action="once",
     category=FutureWarning,
-    message=r"*`get_control_token` is deprecated.*",
+    message=r".*`get_control_token` is deprecated.*",
 )
 
 


src/mistral_common/tokens/tokenizers/sentencepiece.py
```diff
@@ -126,9 +128,14 @@ def eos_id(self) -> int:
         r"""The end of sentence token id."""
         return self._model.eos_id()  # type: ignore
 
-    @cached_property
-    def _control_tokens(self) -> set[int]:
-        return {tok for tok in range(self.n_words) if self._model.IsControl(tok)}
+    def is_special(self, token: int | np.integer | str) -> bool:
```
juliendenize, line 142:
```suggestion
    def is_special_token(self, token: int | np.integer | str) -> bool:
        """Return `True` if the passed `token` is special."""
```

Nit I would have made it explicit that the method is related to tokens directly on the name, but the signature in it self is clear. Also could you add docstring ?

src/mistral_common/tokens/tokenizers/tekken.py
```diff
@@ -203,6 +204,7 @@ def __init__(
         self._audio_config = audio_config
 
         self._all_special_tokens = special_tokens
+        self._special_token_ids = {t["rank"] for t in special_tokens}
         self._special_tokens_reverse_vocab = {t["token_str"]: t["rank"] for t in special_tokens}
         self._vocab = [self.id_to_piece(i) for i in range(vocab_size)]
```
juliendenize, line 216: These are generally accessed in certains ways by vLLM or Transformers integrations.

Can be done later but we should add getters.

src/mistral_common/tokens/tokenizers/tekken.py
```diff
@@ -433,13 +435,25 @@ def is_byte(self, token_id: int) -> bool:
...
 
-    def get_control_token(self, s: str) -> int:
+    def get_special_token(self, s: str) -> int:
         r"""Get the token id of a control token."""
         if s in self._special_tokens_reverse_vocab:
             return self._special_tokens_reverse_vocab[s]
         else:
             raise ValueError(f"Unknown control token {s}")
 
+    def is_special(self, token: int | np.integer | str) -> bool:
```
juliendenize, line 452: Same comment as for spm:
- (optional) would have named it `is_special_token`
- add docstring please

```suggestion
    def is_special_token(self, token: int | np.integer | str) -> bool:
        """Return `True` if the passed `token` is special."""
 ```

File: src/mistral_common/tokens/tokenizers/tekken.py (hunk 2 above)
Comment (patrickvonplaten) [reply src/mistral_common/tokens/tokenizers/tekken.py:216]: Would like to try to not have to access them actually, but ok let's see

File: src/mistral_common/tokens/tokenizers/sentencepiece.py (hunk 1 above)
Comment (patrickvonplaten) [reply src/mistral_common/tokens/tokenizers/sentencepiece.py:142]: Can't rename function here anymore as it'd won't be sync with other codebase

Comment (patrickvonplaten) [inline src/mistral_common/tokens/tokenizers/sentencepiece.py:142]:
```suggestion
    def is_special(self, token: int | np.integer | str) -> bool:
           """Return `True` if the passed `token` is a special token."""
```

src/mistral_common/tokens/tokenizers/tekken.py
patrickvonplaten, line 452:
```suggestion
    def is_special(self, token: int | np.integer | str) -> bool:
           """Return `True` if the passed `token` is a special token."""
```

src/mistral_common/tokens/tokenizers/tekken.py
```diff
@@ -433,13 +435,26 @@ def is_byte(self, token_id: int) -> bool:
         r"""Check if a token id is a byte token."""
         return 0 <= token_id - self.num_special_tokens < 256
 
-    def get_control_token(self, s: str) -> int:
+    def get_special_token(self, s: str) -> int:
         r"""Get the token id of a control token."""
```
patrickvonplaten:
```suggestion
        r"""Get the token id of a special token."""
```

src/mistral_common/tokens/tokenizers/sentencepiece.py
```diff
@@ -103,7 +105,7 @@ def version(self) -> TokenizerVersion:
         r"""The version of the tokenizer."""
         return self._version
 
-    def get_control_token(self, s: str) -> int:
+    def get_special_token(self, s: str) -> int:
         r"""Get the control token for the given string."""
```
patrickvonplaten:
```suggestion
        r"""Get the special token for the given string."""
```

src/mistral_common/tokens/tokenizers/base.py
```diff
@@ -241,9 +241,13 @@ def decode(self, tokens: list[int], special_token_policy: SpecialTokenPolicy | N
         """
 
     @abstractmethod
-    def get_control_token(self, s: str) -> int:
```
juliendenize, line 244: Can we not delete and deprecate ?

src/mistral_common/tokens/tokenizers/tekken.py
```diff
@@ -433,13 +435,26 @@ def is_byte(self, token_id: int) -> bool:
         r"""Check if a token id is a byte token."""
         return 0 <= token_id - self.num_special_tokens < 256
 
-    def get_control_token(self, s: str) -> int:
```
juliendenize, line 436: can we not delete and deprecate ?

juliendenize (changes requested):
Think we need to check before that these methods are not used in vLLM and Transformers.

Edit:
Yes they are used 
vLLM: https://github.com/vllm-project/vllm/blob/main/vllm/tokenizers/mistral.py#L539
https://github.com/vllm-project/vllm/blob/main/vllm/tokenizers/mistral.py#L299
Transformers:
https://github.com/huggingface/transformers/blob/6217adc6c8f0be7b5374e6a46129ad2214e4c6ed/src/transformers/tokenization_mistral_common.py#L609

src/mistral_common/tokens/tokenizers/tekken.py
```diff
@@ -433,13 +435,26 @@ def is_byte(self, token_id: int) -> bool:
...
+    def is_special(self, token: int | np.integer | str) -> bool:
+        """Return `True` if the passed `token` is a special token."""
+        if isinstance(token, (int, np.integer)):
+            return token in self._special_token_ids
+        elif isinstance(token, str):
+            return token in self._special_tokens_reverse_vocab
+        else:
+            raise TypeError(f"Expected int or str, got {type(token).__name__}")
+
+    def get_control_token(self, s: str) -> int:
```
patrickvonplaten, line 461: To not break anything

src/mistral_common/tokens/tokenizers/sentencepiece.py
```diff
@@ -103,8 +105,8 @@ def version(self) -> TokenizerVersion:
         r"""The version of the tokenizer."""
         return self._version
 
-    def get_control_token(self, s: str) -> int:
-        r"""Get the control token for the given string."""
+    def get_special_token(self, s: str) -> int:
+        r"""Get the special token for the given string."""
         return self._model.piece_to_id(s)  # type: ignore
```
patrickvonplaten, line 117:
```suggestion
        return self._model.piece_to_id(s)  # type: ignore
        
    def get_control_token(self, s: str) -> int:
        warnings.warn("`get_control_token` is deprecated. Use `get_special_token` instead.", FutureWarning)
        return self.get_special_token(s)
```

src/mistral_common/tokens/tokenizers/tekken.py
juliendenize: could you warn only once ?

src/mistral_common/tokens/tokenizers/tekken.py
juliendenize: sorry was handled below closing

src/mistral_common/tokens/tokenizers/sentencepiece.py
juliendenize: yep thanks overlooked on second reading -_-'

src/mistral_common/tokens/tokenizers/base.py
juliendenize: here it doesn't matter

juliendenize (approved):
LGTM thanks !

PR #172, opened 2025-12-21
Add new token logic asrstr

patrickvonplaten:
Add some new logic for audio

### DIFF
--- a/src/mistral_common/tokens/tokenizers/audio.py
+++ b/src/mistral_common/tokens/tokenizers/audio.py
@@ -270,7 +270,7 @@ class AudioEncoder:
 
     def _encode_audio_chunk(self, content: AudioChunk) -> AudioEncoding:
         audio = Audio.from_raw_audio(content.input_audio)
-        return self.encode_audio(audio, False)
+        return self.encode_audio(audio, is_online_streaming=False)
 
     def _encode_audio_url_chunk(self, content: AudioURLChunk) -> AudioEncoding:
         url_type = content.get_url_type()


src/mistral_common/tokens/tokenizers/audio.py
```diff
@@ -9,6 +10,22 @@
 
 logger = logging.getLogger(__name__)
 
+# for offline streaming we're encoding the whole audio at once.
+# Because the model is delayed by <transcription_delay_ms> + word_length
+# We must always add a buffer of max world length to the audio in the end
+# For now we assume that there is no word that requires more than 10 tokens (0.8s)
+# which it might it practice but it won't affect any eval results
```
```suggestion
# For offline streaming we're encoding the whole audio at once.
# Because the model is delayed by <transcription_delay_ms> + word_length
# we must always add a buffer of max world length to the audio in the end.
# For now we assume that there is no word that requires more than 10 tokens (0.8s)
# which might not be the case in practice but it won't affect any eval results.
```

src/mistral_common/tokens/tokenizers/audio.py
```diff
@@ -9,6 +10,22 @@
...
+OFFLINE_STREAMING_BUFFER_TOKENS = 10
+
+
+class TranscriptionFormat(str, Enum):
+    """Transcription format
+    Should be set by the tokenizer for correct encoding.
```
```suggestion
    r"""Transcription format.

    Should be set by the tokenizer for correct encoding.
    
    Attributes:
    - INSTRUCT: The instruct format.
    - STREAMING: The streaming format.
```

src/mistral_common/tokens/tokenizers/audio.py
```diff
@@ -52,6 +69,12 @@ class AudioConfig:
     # Whether to pad an audio into multiples of chunk_length_s seconds
     chunk_length_s: float | None = None
 
+    # delay between the audio stream and text stream
+    transcription_delay_ms: float | None = None
+
+    # If we're in streaming or non-streaming
+    transcription_format: TranscriptionFormat | None = TranscriptionFormat.INSTRUCT
```
juliendenize: If I understand correctly, for `None` we also assume that it is an instruct mode right ? I'd presume it is for backward compatibility ? Is it necessary ? if we set a default value, the ones that didn't define it will have INSTRUCT right ?

If it's necessary, is it possible in `__post_init__` to set None to TranscriptionFormat.INSTRUCT ?

src/mistral_common/tokens/tokenizers/audio.py
```diff
@@ -62,16 +85,54 @@ def __post_init__(self) -> None:
...
+        else:
+            audio_len = audio_len // self.encoding_config.hop_length
+
+        return math.ceil(audio_len / self.audio_length_per_tok)
+
+    @property
+    def num_delay_tokens(self) -> int:
+        assert self.is_streaming, f"Can't call num_delay_tokens if {self.is_streaming=}."
+        # streaming pad tokens
+        assert self.transcription_delay_ms is not None
```
```suggestion
        assert self.transcription_delay_ms is not None, f"Can't call num_delay_tokens if {self.transcription_delay_ms=}."
```

src/mistral_common/tokens/tokenizers/audio.py
```diff
@@ -118,25 +179,49 @@ def __init__(self, audio_config: AudioConfig, special_ids: SpecialAudioIDs) -> N
...
+        mult_of = self.audio_config.raw_audio_length_per_tok
+        pad = int((mult_of - (num_samples % mult_of)) % mult_of)
+
+        if not is_online:
+            # in offline streaming we're appending an extra pad to simulate
+            # a whole streaming session
+
+            #  then add delay tokens + BOS token + buffer approx
+            _extra_pad_tokens = (self.audio_config.num_delay_tokens + 1) + OFFLINE_STREAMING_BUFFER_TOKENS
+            extra_pad_samples = int(mult_of * _extra_pad_tokens)
```
juliendenize, line 226: nit: why private variable ? personal pref i don't like them
```suggestion
            extra_pad_tokens = (self.audio_config.num_delay_tokens + 1) + OFFLINE_STREAMING_BUFFER_TOKENS
            extra_pad_samples = int(mult_of * extra_pad_tokens)
```

src/mistral_common/tokens/tokenizers/audio.py
```diff
@@ -178,7 +263,7 @@ def _encode_audio(self, audio: Audio) -> AudioEncoding:
 
     def _encode_audio_chunk(self, content: AudioChunk) -> AudioEncoding:
         audio = Audio.from_raw_audio(content.input_audio)
-        return self._encode_audio(audio)
+        return self.encode_audio(audio, False)
```
juliendenize: nit: to make call consistent with below

```suggestion
        return self.encode_audio(audio, is_online_streaming=False)
```

src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -944,7 +948,21 @@ def encode_transcription(self, request: TranscriptionRequest) -> Tokenized:
         Returns:
             Tokenized: The tokenized representation of the audio data, including processed audio and tokens
         """
+        assert self.audio_encoder is not None, f"Audio encoder must be defined, got {self.audio_encoder=}"
+        if self.audio_encoder.audio_config.transcription_format == TranscriptionFormat.INSTRUCT:
+            return self._encode_instruct_transcription(request)
+        elif self.audio_encoder.audio_config.transcription_format == TranscriptionFormat.STREAMING:
+            return self._encode_streaming_transcription(request)
```
juliendenize, line 955: This matches what I said above, audio_config should not allow None right ?

src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -963,6 +981,43 @@ def encode_transcription(self, request: TranscriptionRequest) -> Tokenized:
         tokens.append(self.TRANSCRIBE)
         return Tokenized(tokens=tokens, text=self.tokenizer._to_string(tokens), audios=audio)
 
+    def _encode_audio(self, audio: str | bytes, is_online_streaming: bool) -> Tokenized:
+        assert self.audio_encoder is not None, (
+            f"Audio encoder must be defined to encode audio, got {self.audio_encoder=}"
+        )
+        _audio = Audio.from_base64(audio) if isinstance(audio, str) else Audio.from_bytes(audio)
+        audio_enc = self.audio_encoder.encode_audio(_audio, is_online_streaming=is_online_streaming)
```
juliendenize, line 989: nit: not huge fan of private variables but it's personal preference 

```suggestion
        audio_obj = Audio.from_base64(audio) if isinstance(audio, str) else Audio.from_bytes(audio)
        audio_enc = self.audio_encoder.encode_audio(audio_obj, is_online_streaming=is_online_streaming)
```

tests/test_tekken.py
```diff
@@ -18,7 +18,7 @@
 )
 
 
-def _quick_vocab(extra_toks: Sequence[bytes] = ()) -> list[TokenInfo]:
+def quick_vocab(extra_toks: Sequence[bytes] = ()) -> list[TokenInfo]:
```
juliendenize, line 21: Thanks for renaming

tests/test_tokenizer_v7_audio_streaming.py
```diff
@@ -0,0 +1,173 @@
...
+        except IndexError as exc:
+            raise ValueError(f"special_tokens is missing index {idx}") from exc
+
+        existing = token_entry.get("token_str")
+        placeholder = f"<SPECIAL_{idx}>"
+        if existing not in (None, placeholder):
+            raise ValueError(f"special_tokens[{idx}] already set to {existing}")
+
+        token_entry["token_str"] = token
+        token_entry["is_control"] = is_control
```
juliendenize, line 56: Could be subsequent pr but I think this and functions imported here:
```
from tests.test_tekken import get_special_tokens, quick_vocab
```
should be moved to some utils file.

src/mistral_common/tokens/tokenizers/audio.py
patrickvonplaten: Hmm yeah guess that's style related. Think private variables are nice because they indicate that they only used locally

src/mistral_common/tokens/tokenizers/audio.py
patrickvonplaten: Hmm do be honestly mypy should catch if there is a problem no?

patrickvonplaten: But ok

tests/test_tokenizer_v7_audio_streaming.py
patrickvonplaten: Yeah won't do in this PR though

src/mistral_common/tokens/tokenizers/instruct.py
patrickvonplaten: yeah agree

PR #176, opened 2026-01-20
Enhance accessibility

juliendenize:
This PR introduces some minor improvements to the tokenizer and validation system to ease integrations:
1. Str Enums: Converted ValidationMode, SpecialTokenPolicy, and TokenizerVersion to use str, Enum
2. New Tokenizer Properties: Added special_ids and num_special_tokens properties to all tokenizer implementations for easier access to special token information
3. Enhanced Accessibility: Added version and mode properties to InstructTokenizer and MistralTokenizer

### DIFF
--- a/src/mistral_common/tokens/tokenizers/base.py
+++ b/src/mistral_common/tokens/tokenizers/base.py
@@ -110,6 +110,7 @@ class SpecialTokenPolicy(str, Enum):
 
     @classmethod
     def _missing_(cls, value: Any) -> Any:
+        # Backward compatibility of int values.
         match value:
             case 0:
                 return SpecialTokenPolicy.IGNORE
--- a/tests/test_tokenizer_base.py
+++ b/tests/test_tokenizer_base.py
@@ -9,6 +9,6 @@ def test_special_token_policy_backward_compatibility() -> None:
     assert SpecialTokenPolicy(2) == SpecialTokenPolicy.RAISE
 
     with pytest.raises(ValueError, match=r"3 is not a valid SpecialTokenPolicy"):
-        SpecialTokenPolicy(3) is None
+        SpecialTokenPolicy(3)
     with pytest.raises(ValueError, match=r"'invalid' is not a valid SpecialTokenPolicy"):
-        SpecialTokenPolicy("invalid") is None
+        SpecialTokenPolicy("invalid")


src/mistral_common/tokens/tokenizers/base.py
```diff
@@ -105,9 +104,20 @@ class SpecialTokenPolicy(int, Enum):
...
 
-    IGNORE = 0
-    KEEP = 1
-    RAISE = 2
+    IGNORE = "ignore"
+    KEEP = "keep"
+    RAISE = "raise"
+
+    @classmethod
+    def _missing_(cls, value: Any) -> Any:
```
patrickvonplaten, line 112: what is this for?

src/mistral_common/tokens/tokenizers/sentencepiece.py
```diff
@@ -156,7 +161,14 @@ def is_special(self, token: int | np.integer | str) -> bool:
...
-        warnings.warn("`_control_tokens` is deprecated. Make use of `is_special` instead.", FutureWarning)
+        warnings.warn(
+            "`_control_tokens` is deprecated. Make use of `is_special` or `special_ids` instead.", FutureWarning
+        )
+        return {tok for tok in range(self.n_words) if self._model.IsControl(tok)}
+
+    @cached_property
+    def special_ids(self) -> set[int]:
+        r"""Ids of the special tokens."""
         return {tok for tok in range(self.n_words) if self._model.IsControl(tok)}
```
patrickvonplaten, line 172: is this correct? For SPM len(control_tokens) == len(special_tokens)?

src/mistral_common/tokens/tokenizers/tekken.py
```diff
@@ -328,6 +328,11 @@ def n_words(self) -> int:
         r"""Vocabulary size of the tokenizer."""
         return self._vocab_size
 
+    @cached_property
+    def special_ids(self) -> set[int]:
```
patrickvonplaten, line 332: where is this property used exactly?

src/mistral_common/tokens/tokenizers/sentencepiece.py
juliendenize: Yes I could actually make self.control_tokens fall back to special_ids and keep the deprecation warning instead.

src/mistral_common/tokens/tokenizers/base.py
juliendenize: If you mean why __missing__, it's to keep backward compatibility before it was 0,1,2

If you mean why the change, then it's just to provide an API that is easier to access for the user without the need of using the enum. Related to that I just pushed a commit to allow str being passed otherwise it would have failed.

src/mistral_common/tokens/tokenizers/tekken.py
juliendenize: Not yet but it will be used by Transformers and vLLM.

tests/test_tokenizer_base.py
```diff
@@ -0,0 +1,14 @@
...
+
+def test_special_token_policy_backward_compatibility() -> None:
+    assert SpecialTokenPolicy(0) == SpecialTokenPolicy.IGNORE
+    assert SpecialTokenPolicy(1) == SpecialTokenPolicy.KEEP
+    assert SpecialTokenPolicy(2) == SpecialTokenPolicy.RAISE
+
+    with pytest.raises(ValueError, match=r"3 is not a valid SpecialTokenPolicy"):
+        SpecialTokenPolicy(3) is None
+    with pytest.raises(ValueError, match=r"'invalid' is not a valid SpecialTokenPolicy"):
+        SpecialTokenPolicy("invalid") is None
```
juliendenize:
```suggestion
        SpecialTokenPolicy(3)
    with pytest.raises(ValueError, match=r"'invalid' is not a valid SpecialTokenPolicy"):
        SpecialTokenPolicy("invalid")
```

src/mistral_common/tokens/tokenizers/base.py
```diff
@@ -105,9 +104,20 @@ class SpecialTokenPolicy(int, Enum):
...
+        match value:
```
juliendenize, line 114:
```suggestion
        # Backward compatibility of int values.
        match value:
```

PR #177, opened 2026-01-22
[Audio] Improve padding for streaming

patrickvonplaten:
Left padding improves results

### DIFF
--- a/src/mistral_common/tokens/tokenizers/audio.py
+++ b/src/mistral_common/tokens/tokenizers/audio.py
@@ -150,7 +150,7 @@ class AudioConfig:
     @property
     def n_right_pad_tokens(self) -> int:
         assert self.is_streaming, f"Can't call n_right_pad_tokens if {self.is_streaming=}."
-        # we need to pad on the rigth to ensure the models transcribes
+        # we need to pad on the right to ensure the models transcribes
         # - the induced delay on the prefill step (num_delay_tokens)
         # - the BOS token (1)
         # - a heuristic that defines a max token length for a single word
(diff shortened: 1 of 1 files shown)


src/mistral_common/tokens/tokenizers/audio.py
```diff
@@ -143,6 +147,25 @@ def audio_length_per_tok(self) -> int:
...
         return int(downsample_factor)
 
+    @property
+    def n_right_pad_tokens(self) -> int:
+        assert self.is_streaming, f"Can't call n_right_pad_tokens if {self.is_streaming=}."
+        # we need to pad on the rigth to ensure the models transcribes
+        # - the induced delay on the prefill step (num_delay_tokens)
+        # - the BOS token (1)
+        # - a heuristic that defines a max token length for a single word
+        #   (OFFLINE_STREAMING_BUFFER_TOKENS)
```
```suggestion
        # we need to pad on the right to ensure the models transcribes
        # - the induced delay on the prefill step (num_delay_tokens)
        # - the BOS token (1)
        # - a heuristic that defines a max token length for a single word
        #   (OFFLINE_STREAMING_BUFFER_TOKENS)
```

src/mistral_common/tokens/tokenizers/audio.py
```diff
@@ -202,8 +225,10 @@ def pad(self, audio_array: np.ndarray, sampling_rate: int) -> np.ndarray:
             next_multiple_of_chunk_frames = self.next_multiple_of_chunk_frames(audio_array.shape[-1], sampling_rate)
             audio_array = np.pad(audio_array, (0, next_multiple_of_chunk_frames - audio_array.shape[-1]))
         elif self.audio_config.is_streaming:
-            pad = self._get_streaming_pad(audio_array.shape[-1])
-            audio_array = np.pad(audio_array, (0, pad))
+            left_pad, right_pad = self._get_streaming_pad(audio_array.shape[-1])
+            # we pad both left & right as this leads to better performance
+            # doesn't seem to work as expected here
```
juliendenize: I don't get what this lines mean with respect to the one just above.
```suggestion
            # doesn't seem to work as expected here
```

patrickvonplaten: ah ups nice catch 😅

patrickvonplaten:
```suggestion
```

PR #178, opened 2026-01-26
Add TokenizerV14

juliendenize:
This PR adds the tokenizer v14 tokenizer.

### DIFF
--- a/src/mistral_common/protocol/instruct/request.py
+++ b/src/mistral_common/protocol/instruct/request.py
@@ -33,6 +33,26 @@ class ResponseFormats(str, Enum):
     json = "json_object"
 
 
+class ReasoningEffort(str, Enum):
+    r"""Enum of the reasoning effort for the model.
+
+    Note:
+        This is used for `ChatCompletionRequest` or `InstructRequest` for tokenizers >= v14
+
+    Attributes:
+        none: No reasoning is done.
+        high: Reasoning should be performed by the model.
+
+    Examples:
+        >>> reasoning_effort = ReasoningEffort.high
+        >>> reasoning_effort = ReasoningEffort.none
+        >>> reasoning_effort = None
+    """
+
+    none = "none"
+    high = "high"
+
+
 class ResponseFormat(MistralBase):
     r"""The format of the response.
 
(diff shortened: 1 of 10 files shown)


src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -1208,3 +1229,51 @@ def encode_think(self, chunk: ThinkChunk) -> list[int]:
...
+        self,
+        tokenizer: Tokenizer,
+        image_encoder: ImageEncoder | None = None,
+        audio_encoder: AudioEncoder | None = None,
+    ) -> None:
+        super().__init__(tokenizer, image_encoder, audio_encoder=audio_encoder)
+        self.BEGIN_MODEL_SETTINGS = self.tokenizer.get_special_token(SpecialTokens.begin_model_settings.value)
+        self.END_MODEL_SETTINGS = self.tokenizer.get_special_token(SpecialTokens.end_model_settings.value)
+
+    def _encode_settings(
```
patrickvonplaten, line 1251: let's make this more general for all settings

src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -431,32 +442,38 @@ def encode_user_message(
...
+        if do_encode_settings and available_tools:
+            tools = [tool.model_dump(exclude={"function": {"strict": True}}) for tool in available_tools]
             tools_json_tokens = self.tokenizer.encode(json.dumps(tools, ensure_ascii=False), bos=False, eos=False)
             tools_tokens = [
                 self.BEGIN_AVAILABLE_TOOLS,
                 *tools_json_tokens,
                 self.END_AVAILABLE_TOOLS,
             ]
 
+        if do_encode_settings:
```
patrickvonplaten, line 466: Interesting that we encode reasoning effort for every user message

patrickvonplaten: I would have assumed that we should only ever encode model settings once per conversation

src/mistral_common/protocol/instruct/request.py
```diff
@@ -76,6 +98,7 @@ class ChatCompletionRequest(BaseCompletionRequest, Generic[ChatMessageType]):
     model: str | None = None
     messages: list[ChatMessageType]
     response_format: ResponseFormat = Field(default_factory=ResponseFormat)
+    reasoning_effort: ReasoningEffort | None = None
```
patrickvonplaten, line 101: I think we should consider instead doing:


```suggestion
    ModelSetting = ReasoningEffort | ....
    ...
    model_settings: list[ModelSetting] | None = None
```

src/mistral_common/tokens/tokenizers/instruct.py
joa-stdn: I think `do_encode_settings` should be only True for the first user message?

PR #184, opened 2026-02-05
Allow System Prompt with Audio for v13

juliendenize:


amosyou (approved):
working e2e on vllm!

tests/validation/test_chat_validation.py
```diff
@@ -395,3 +440,16 @@ def test_multiple_assistant_messages_some_with_tool_calls(self, validator_v13: M
             ],
             continue_final_message=False,
         )
+
+    @pytest.mark.parametrize("audio_fixture", ["audio_chunk", "audio_url_chunk"])
+    def test_audio_with_system_prompt_raises_ok(
```
juliendenize:
```suggestion
    def test_audio_with_system_prompt_ok(
```

src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -1010,6 +1010,7 @@ def _encode_streaming_transcription(self, request: TranscriptionRequest) -> Toke
 
     @classmethod
     def validate_messages(cls, messages: list[UATS]) -> None:
+        r"""Validates that system prompts and audio chunks are not used together in v7."""
```
joa-stdn, line 1048: why can't we validate in validator only and not tokenizer?

src/mistral_common/protocol/instruct/validator.py
```diff
@@ -402,14 +405,40 @@ class MistralRequestValidatorV5(MistralRequestValidatorV3):
 
     This validator allows for both tool calls and content in the assistant message.
 
+    Note:
+        For requests containing audio, this validator ensures that no system prompt are present.
```
patrickvonplaten:
```suggestion
        For requests containing audio, this validator ensures that no system prompt is present.
```

src/mistral_common/protocol/instruct/validator.py
```diff
@@ -402,14 +405,40 @@ class MistralRequestValidatorV5(MistralRequestValidatorV3):
...
     Examples:
         >>> validator = MistralRequestValidatorV5()
     """
 
     _allow_tool_call_and_content: bool = True
 
+    def _validate_system_prompt_and_audio(self, messages: list[UATS]) -> None:
+        r"""Validates that system prompts and audio chunks are not used together in v5."""
+        has_audio = any(
+            isinstance(message, UserMessage)
```
patrickvonplaten: we soon will have audio chunks in the AssistantMessage as well - is this still fine here?

src/mistral_common/protocol/instruct/validator.py
```diff
@@ -402,14 +405,40 @@ class MistralRequestValidatorV5(MistralRequestValidatorV3):
...
+            and isinstance(message.content, list)
+            and any(isinstance(chunk, (AudioChunk, AudioURLChunk)) for chunk in message.content)
+            for message in messages
+        )
+        has_sp = any(isinstance(message, SystemMessage) for message in messages)
+        if has_audio and has_sp:
+            raise ValueError("System messages are not yet allowed when audio is present")
```
patrickvonplaten:
```suggestion
            raise ValueError(f"System messages {...} are not yet allowed when audio is present as in the case for {...}")
```

would be nice to show a better error message here

tests/conftest.py
```diff
@@ -0,0 +1,29 @@
+import numpy as np
+import pytest
+
+from mistral_common.audio import Audio
+from mistral_common.protocol.instruct.chunk import AudioChunk, AudioURL, AudioURLChunk, RawAudio
+
+
+@pytest.fixture(scope="session")
+def audio_chunk() -> AudioChunk:
```
patrickvonplaten: can we not put this in a super generic `conftest.py` but instead directly in the test_chat_validation.py file?

tests/validation/test_chat_validation.py
```diff
@@ -262,7 +269,45 @@ def test_too_many_tool_messages(self, validator: MistralRequestValidator) -> Non
             )
 
 
-class TestChatValidationV6:
+class TestChatValidationV5:
```
patrickvonplaten, line 283: tests look nice

tests/conftest.py
patrickvonplaten: Hmm any chance we can put this rather in audio_fixtures.py file and import it from there?

src/mistral_common/tokens/tokenizers/instruct.py
juliendenize: It's for backward compatibility it was already there before so I leave it. I could deprecate if you prefer

src/mistral_common/protocol/instruct/validator.py
juliendenize: not sure to fully understand the question here:
- do you mean that i should already cover assistant messages ?
- are you afraid this would be triggered (answer is no we check for user messages)

src/mistral_common/protocol/instruct/validator.py
juliendenize: Improved a bit by giving index locations of the messages, lmk if it's good now

tests/conftest.py
juliendenize: Created a fixture folder for just audio now but it'd be nice to also clean up a bit the files that redefine a lot of fixtures for tokenizers later. I still need to redefine the fixtures on top of the file, this is because import fixtures is unclean especially when using `getfixturevalue` as ruff complain about imports not used.

is it good enough ?

tests/test_tokenizer_v13.py
```diff
@@ -16,13 +18,25 @@
...
 
 
+@pytest.fixture(scope="module")
+def audio_chunk() -> AudioChunk:
+    return get_dummy_audio_chunk()
+
+
+@pytest.fixture(scope="module")
+def audio_url_chunk() -> AudioURLChunk:
+    return get_dummy_audio_url_chunk()
```
patrickvonplaten, line 37: can we not define these fixtures in `fixtures/audio.py` and import them?

src/mistral_common/protocol/instruct/validator.py
patrickvonplaten: My question is that we'll soon have a TTS model probs based on V7 and in this case one could have a [SystemPrompt(audio), UserMessage(text), Assistant(audio), UserMessage(text)] and it would'nt fail - is this ok / expected? 

(probably a bit of a nit)

juliendenize: I think we'd need to check internally, if it's not then we'll need to add indeed AssistantMessage in the check. But I think we can wait for the actual release to add it as it isn't breaking backward compatibility and mistral-common does not support AudioChunk for now outside of user message.

tests/test_tokenizer_v13.py
juliendenize: You can but it's breaking pytest design. Not so sure about underlying consequences of doing that so I think it's fine to do it this way and maybe later reconsider.

PR #190, opened 2026-02-21
ci: add dependency caching and split lint from test matrix

framsouza:
**Dependency caching**: `setup-uv` has built-in cache support keyed on `uv.lock` but it wasn't enabled in the test or docs jobs. Added `enable-cache: true` and `cache-dependency-glob: "uv.lock"` to all jobs. The docs workflow already had its own caching setup so this makes things consistent.

**Lint runs once**: ruff and mypy were running inside the 4-version test matrix, which means they were running 4 times and producing the same result every time. Moved them into a separate `lint` job on Python 3.12. The test job now has `needs: lint` so if linting fails nothing else starts.

**Coverage artifact**: pytest was generating `coverage.xml` on every run but the file was never saved anywhere. Added an artifact upload so it's accessible from the Actions tab per Python version.

Each PR commit currently triggers 4 redundant package downloads and runs linting 4 times for no gain. With caching and the lint split, subsequent runs skip the downloads entirely and linting consumes one job instead of four. This translates directly to faster feedback and fewer GitHub Actions compute minutes burned, which at scale has a real cost impact.

## Testing

Tested locally with `act pull_request`. All jobs pass. 

```
| ================================ tests coverage ================================
| _______________ coverage: platform linux, python 3.12.3-final-0 ________________
| 
| Coverage XML written to file coverage.xml
| ================ 506 passed, 16 skipped, 38 warnings in 57.61s =================
```

**NOTE**: `test_from_url` will show as failing on this PR in CI until https://github.com/mistralai/mistral-common/pull/188  is merged into main first.


.github/workflows/lint_build_test.yaml
```diff
@@ -20,30 +20,19 @@ concurrency:
   cancel-in-progress: ${{ github.event.pull_request.number != null }}
 
 jobs:
-  build_lint_test:
```
juliendenize, line 23: Think we can keep it this way

.github/workflows/lint_build_test.yaml
```diff
@@ -20,30 +20,19 @@ concurrency:
...
         uses: astral-sh/setup-uv@d4b2f3b6ecc6e67c4457f6d3e41ec42d3d0fcb86 # v5
         with:
-          python-version: ${{ matrix.python-version }}
-      
-      - name: Check uv lock
-        run: |
-          uv lock --check
+          python-version: "3.12"
+          enable-cache: true
+          cache-dependency-glob: "uv.lock"
```
juliendenize: This is already handled by default settings AFAIK

.github/workflows/lint_build_test.yaml
```diff
@@ -65,32 +54,74 @@ jobs:
...
         run: |
           uv run pytest --cov=mistral_common . --cov-report "xml:coverage.xml"
-      
+
+      # Save coverage report as artifact so it's not discarded after the job ends
+      - name: Upload coverage
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-${{ matrix.python-version }}
+          path: coverage.xml
```
juliendenize: no need to store coverage as we don't use it :smile:

.github/workflows/lint_build_test.yaml
```diff
@@ -60,37 +48,76 @@ jobs:
...
+    needs: lint
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # Support all LTS versions of Python
+        python-version: [
+          "3.10",
+          "3.11",
+          "3.12",
+          "3.13"
```
juliendenize:
```suggestion
          "3.13",
          "3.14",
```

juliendenize (changes requested):
There are still adjustments to make as now we support 3.14 and i don't think we should split linting from tests as we want the linter to work on all versions.

.github/workflows/lint_build_test.yaml
framsouza: That was removed here https://github.com/mistralai/mistral-common/pull/190/changes/b43ee3d2fc6df301857131e73ea18f453a53e165

.github/workflows/lint_build_test.yaml
framsouza: indeed, nobody is realistically going to download this artifacts to review coverage

PR #205, opened 2026-03-30
Allow multimodal tool messages for improved agentic experience

simonwm:
Thanks for all the effort to make cutting edge Mistral models compatible with vllm and others, I really appreciate this!

Maybe this PR can improve it a bit further. It implements compatibility with *multimodal tool* messages. More than implementing something new, it merely relaxes some constraints for tools which are not there for user messages. I have to admit that I don't quite get why not all messages can just be multimodal by default...

Right now the PR works nicely (e.g., if applied in a container derived from mistralllm/vllm-ms4:latest), but probably does not follow all the conventions in the repository, e.g. probably breaks the versioning scheme in src/mistral_common/tokens/tokenizers/instruct.py and just changes code from the old InstructTokenizerV13...

Additional context:

I stumbled over this issue when I was experimenting with a langchain deepagent who should create some PNGs and then look at them to verify whether it did the right thing. Although the deepagent framework implements a `read_file` tool which returns images as multimodal content blocks (https://docs.langchain.com/oss/python/deepagents/harness#virtual-filesystem-access), the workflow did not work using Mistral models hosted in different ways.

Something else than text is also not allowed in the OpenAI specs either (https://github.com/openai/openai-python/blob/58184ad545ee2abd98e171ee09766f259d7f38cd/src/openai/types/chat/chat_completion_tool_message_param.py#L14), so for OpenAI compatibility (in the sense of not allowing something that they don't) one would need to upstream this change there as well.


juliendenize:
Hey thanks for the PR, i'm building on the side PRs to expand the chunks allowed to messages. As some older models didn't have access to this, i'm obligated to do more checks etc so it would be quite different to what you're making here.
The result will however be the same as long as you use recent models (using v15+ tokenizers).

simonwm:
Thanks for looking into this! Sounds like a good plan. Looking forward to the release!

juliendenize:
Fixed by #241

PR #214, opened 2026-04-24
Add agentic validation mode

juliendenize:
This PR adds agentic validation mode that allows tool interruption followed by user message. This allows usage of coding agents through vLLM serving for example as soon as we add this behavior to vLLM.


juliendenize:
#218 should be enough to unlock agentic :)

PR #215, opened 2026-04-24
Add default validation mode

juliendenize:
This PR brings the possibility to initialize a ValidationMode for MistralTokenizer from a default stored in tekken.json.


juliendenize:
#218 should be enough to unlock agentic :)

PR #217, opened 2026-04-24
Make from_openai methods lenient by silently dropping unsupported fields

juliendenize:
## Summary

- Simplify `from_openai` usage by silently dropping unsupported fields instead of raising errors
- Remove hard-to-maintain OpenAI field allowlist that broke on SDK updates
- Keep strict validation on direct model construction

This should ease integrations in Transformers and vLLM.


src/mistral_common/protocol/instruct/request.py
```diff
@@ -383,14 +369,16 @@ def from_openai(
...
-        _check_openai_fields_names(valid_fields, set(kwargs.keys()))
+        filtered_kwargs = {k: v for k, v in kwargs.items() if k in valid_fields}
 
         converted_messages: list[ChatMessage] = convert_openai_messages(messages)
 
         converted_tools = convert_openai_tools(tools) if tools is not None else None
 
-        model_settings = ModelSettings.from_openai(**{k: v for k, v in kwargs.items() if k in model_settings_fields})
-        other_kwargs = {k: v for k, v in kwargs.items() if k not in model_settings_fields}
+        model_settings = ModelSettings.from_openai(
```
joa-stdn: I've never understood why we need to_openai and from_openai for this as InstructRequest is private and only for tokenization (seems error prone to not know the tokenizer here)?

juliendenize: you're right, i'll remove these methods for model_settings.

patrickvonplaten (approved):
Lacking a bit of context here & generally bit indifferent whether we should have a "allow list" or not so approving

joa-stdn (approved):
nice!

PR #218, opened 2026-04-25
Allow user message after tool

juliendenize:
This PR allows user message after tool message.


joa-stdn (approved):
super nice thanks a lot!

PR #221, opened 2026-04-29
Improve from openai methods

juliendenize:
This PR does:
- support to multi-turn thinking inside mistral-common from third-party complying with OpenAI API such as vLLM that put reasoning inside a `reasoning` field
- Convert tools with None fields
- Remove mutation of dicts inside from_openai methods.

### DIFF
--- a/src/mistral_common/protocol/instruct/messages.py
+++ b/src/mistral_common/protocol/instruct/messages.py
@@ -180,7 +180,8 @@ class AssistantMessage(BaseMessage):
             case _, None:
                 thinking = reasoning_content
             case _, _:
-                assert reasoning_content == reasoning, "`reasoning_content` and `reasoning` should be equal."
+                if reasoning_content != reasoning:
+                    raise ValueError("`reasoning_content` and `reasoning` should be equal.")
                 thinking = reasoning
 
         if thinking is not None:
--- a/src/mistral_common/protocol/instruct/messages.py
+++ b/src/mistral_common/protocol/instruct/messages.py
@@ -169,6 +169,16 @@ class AssistantMessage(BaseMessage):
         else:
             raise ValueError(f"Unknown content type: {type(openai_content)}")
 
+        reasoning_content = openai_message.get("reasoning", openai_message.get("reasoning_content", None))
+        if reasoning_content is not None:
+            reasoning_chunk = ThinkChunk(thinking=reasoning_content, closed=True)
+            if isinstance(content, str):
+                content = [reasoning_chunk, TextChunk(text=content)]
+            elif content is None:
+                content = [reasoning_chunk]
+            else:
+                content.insert(0, reasoning_chunk)
+
         return cls.model_validate(
             {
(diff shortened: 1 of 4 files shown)


patrickvonplaten (approved):

src/mistral_common/protocol/instruct/messages.py
```diff
@@ -169,6 +169,16 @@ def from_openai(cls, openai_message: dict[str, Any]) -> "AssistantMessage":
         else:
             raise ValueError(f"Unknown content type: {type(openai_content)}")
 
+        reasoning_content = openai_message.get("reasoning", openai_message.get("reasoning_content", None))
+        if reasoning_content is not None:
+            reasoning_chunk = ThinkChunk(thinking=reasoning_content, closed=True)
```
joa-stdn:
```suggestion
            reasoning_chunk = ThinkChunk(thinking=reasoning_content)
```
no? I don't think we should spread the use of closed too much?

src/mistral_common/protocol/instruct/messages.py
```diff
@@ -169,6 +169,16 @@ def from_openai(cls, openai_message: dict[str, Any]) -> "AssistantMessage":
         else:
             raise ValueError(f"Unknown content type: {type(openai_content)}")
 
+        reasoning_content = openai_message.get("reasoning", openai_message.get("reasoning_content", None))
```
joa-stdn: should we add assert that if both "reasoning" and "reasoning_content" exist they are the same?

juliendenize: yes good point

src/mistral_common/protocol/instruct/messages.py
juliendenize: It's just explicitly marked closed rather than implicitly

src/mistral_common/protocol/instruct/tool_calls.py
```diff
@@ -174,6 +190,8 @@ def validate_arguments(cls, v: str | dict[str, Any]) -> str:
         """
         if isinstance(v, dict):
             return json.dumps(v)
+        elif v is None:
+            return json.dumps({})
```
joa-stdn:
```suggestion
            return "{}"
```
is the same no?

src/mistral_common/protocol/instruct/tool_calls.py
```diff
@@ -142,6 +142,22 @@ def to_openai(self) -> dict[str, Any]:
...
+
+        Args:
+            openai_tool: The OpenAI tool definition.
+
+        Returns:
+            The Mistral tool.
+        """
+        openai_tool = openai_tool.copy()
+        if function := openai_tool.get("function"):
+            function = Function._filter_cls_fields(function)
```
joa-stdn: Do you think we should make Function from_openai work instead?

juliendenize: yes seems like a good idea.

PR #223, opened 2026-05-06
Add to_openai reasoning format for AssistantMessage

juliendenize:
This PR introduces the support of multiple conversion of thinking chunks to various format in order to be compliant with inference backends.

For now, our format is the default but it will be changed to vLLM in the future (reasoning field).

### DIFF
--- a/src/mistral_common/protocol/instruct/messages.py
+++ b/src/mistral_common/protocol/instruct/messages.py
@@ -187,7 +187,6 @@ class AssistantMessage(BaseMessage):
             if effective_format == OpenAIReasoningField.thinking:
                 out_dict["content"] = [chunk.to_openai() for chunk in self.content]
             else:
-                # Find the split point between leading ThinkChunks and remaining content.
                 split_idx = 0
                 for chunk in self.content:
                     if isinstance(chunk, ThinkChunk):
--- a/src/mistral_common/protocol/instruct/messages.py
+++ b/src/mistral_common/protocol/instruct/messages.py
@@ -2,7 +2,7 @@ import warnings
 from enum import Enum
 from typing import Any, Literal, TypeVar
 
-from pydantic import Field, field_validator
+from pydantic import Field
 from typing_extensions import Annotated, TypeAlias
 
 from mistral_common.base import MistralBase
@@ -154,25 +154,6 @@ class AssistantMessage(BaseMessage):
     tool_calls: list[ToolCall] | None = None
     prefix: bool = False
 
-    @field_validator("content")
-    @classmethod
-    def _validate_thinking_chunks_are_leading(
-        cls, content: str | list[TextChunk | ThinkChunk] | None
-    ) -> str | list[TextChunk | ThinkChunk] | None:
-        """Validates that all ThinkChunks are contiguous and at the start of the content list."""
-        if not isinstance(content, list):
-            return content
-        seen_non_think = False
-        for chunk in content:
-            if isinstance(chunk, ThinkChunk):
-                if seen_non_think:
-                    raise InvalidAssistantMessageException(
-                        "ThinkChunks must be leading: all ThinkChunks must appear before any other content chunk."
-                    )
-            else:
-                seen_non_think = True
-        return content
-
     def to_openai(
         self,
         convert_thinking_format: OpenAIReasoningField | None = None,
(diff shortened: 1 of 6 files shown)


src/mistral_common/protocol/instruct/messages.py
```diff
@@ -132,17 +155,71 @@ class AssistantMessage(BaseMessage):
...
+                warnings.warn(
+                    "`convert_thinking_format` defaults to 'thinking' but will change to 'reasoning' "
+                    "in 1.13.0. Pass `convert_thinking_format` explicitly to silence this warning.",
+                    FutureWarning,
+                    stacklevel=2,
+                )
+
+            effective_format = convert_thinking_format or OpenAIReasoningField.thinking
+
+            if effective_format == OpenAIReasoningField.thinking:
```
joa-stdn: maybe clearer to call it thinking_chunk? not sure though, since it's a bit different than thinking_blocks

src/mistral_common/protocol/instruct/messages.py
```diff
@@ -132,17 +155,71 @@ class AssistantMessage(BaseMessage):
...
+                for chunk in self.content:
+                    if isinstance(chunk, ThinkChunk):
+                        split_idx += 1
+                    else:
+                        break
+
+                leading_thinks = self.content[:split_idx]
+                remaining = self.content[split_idx:]
+
+                if leading_thinks:
```
joa-stdn: honestly we should just assert thinking is only first chunk no? Doesn't really make sense otherwise for us at least

src/mistral_common/protocol/instruct/request.py
```diff
@@ -311,7 +328,10 @@ def to_openai(self, **kwargs: Any) -> dict[str, list[dict[str, Any]]]:
             openai_messages.append({"role": "system", "content": self.system_prompt})
 
         for message in self.messages:
-            openai_messages.append(message.to_openai())
+            if isinstance(message, AssistantMessage):
```
joa-stdn, line 331: still don't understand why InstructRequest has to_openai, it's a private class for tokenization IMO but maybe unrelated

juliendenize: i'll probably remove indeed, but i'd take care of it in another pr to deprecate before deletion.

src/mistral_common/protocol/instruct/messages.py
```diff
@@ -133,17 +154,66 @@ class AssistantMessage(BaseMessage):
     tool_calls: list[ToolCall] | None = None
     prefix: bool = False
 
-    def to_openai(self) -> dict[str, Any]:
-        r"""Converts the message to the OpenAI format."""
-        out_dict: dict[str, str | list[dict[str, str | dict[str, Any]]]] = {
+    def to_openai(
+        self,
+        convert_thinking_format: OpenAIReasoningField | None = None,
```
joa-stdn: weird to have `convert_thinking_format` that is `OpenAIReasoningField` (naming) no?

tests/test_tokenizer_v13.py
```diff
@@ -278,15 +278,15 @@ def test_encode_think_chunk(v13_tekkenizer_think: InstructTokenizerV13) -> None:
             "A1",
         ),
         (
-            AssistantMessage(content=[TextChunk(text="A1"), ThinkChunk(thinking="T1")]),
-            "A1[THINK]T1[/THINK]",
+            AssistantMessage(content=[ThinkChunk(thinking="T1"), TextChunk(text="A1")]),
```
joa-stdn, line 281: nice catch!

src/mistral_common/protocol/instruct/messages.py
```diff
@@ -133,17 +154,66 @@ class AssistantMessage(BaseMessage):
...
         }
         if self.content is None:
             pass
         elif isinstance(self.content, str):
             out_dict["content"] = self.content
         else:
-            out_dict["content"] = [chunk.to_openai() for chunk in self.content]
+            split_idx = 0
+            seen_non_think = False
+            for chunk in self.content:
```
joa-stdn: hmm why not but I'm pretty sure we could do the same function with fewer lines no?

joa-stdn (approved):
could be nice to simplify the to_openai but otherwise super nice!

PR #224, opened 2026-05-06
Raise multiple format of reasoning for from_openai

juliendenize:
Before reasoning and thinking chunks in the same message were supported.

This change will prevent that. This is to avoid duplication that is not easy to deduplicate.


joa-stdn (approved):
Nice!

PR #226, opened 2026-05-17
Preserve zero OpenAI seed in chat request conversion

pragnyanramtha:
## Root cause

`ChatCompletionRequest.from_openai()` translated OpenAI's `seed` argument with `kwargs.pop("seed", None) or kwargs.pop("random_seed", None)`. Because `0` is falsey, a valid `seed=0` was discarded and converted to `random_seed=None`.

## Change

Preserve `seed=0` by checking explicitly for `None` before falling back to `random_seed`.

## Validation

- `uv run pytest tests/validation/test_request.py`
- `uv run ruff check src/mistral_common/protocol/instruct/request.py tests/validation/test_request.py`


juliendenize (approved):
Makes sense thanks !

Issue #229, opened 2026-05-27
[BUG: Model name to tokenizer mapping is outdated


copierrj:
### Python -VV

```shell
Python 3.12.13 (main, May 19 2026, 23:48:44) [GCC 14.2.0]
```

### Pip Freeze

```shell
annotated-doc==0.0.4
annotated-types==0.7.0
anyio==4.13.0
attrs==26.1.0
certifi==2026.5.20
charset-normalizer==3.4.7
click==8.4.1
fastapi==0.136.3
filelock==3.29.0
fsspec==2026.4.0
h11==0.16.0
hf-xet==1.5.0
httpcore==1.0.9
httpx==0.28.1
huggingface_hub==1.16.4
idna==3.16
jsonschema==4.26.0
jsonschema-specifications==2025.9.1
markdown-it-py==4.2.0
mdurl==0.1.2
mistral_common==1.11.2
numpy==2.3.5
packaging==26.2
pillow==12.2.0
pycountry==26.2.16
pydantic==2.13.4
pydantic-extra-types==2.11.1
pydantic_core==2.46.4
Pygments==2.20.0
PyYAML==6.0.3
referencing==0.37.0
regex==2026.5.9
requests==2.34.2
rich==15.0.0
rpds-py==0.30.0
sentencepiece==0.2.1
shellingham==1.5.4
starlette==1.1.0
tiktoken==0.13.0
tqdm==4.67.3
typer==0.25.1
typing-inspection==0.4.2
typing_extensions==4.15.0
urllib3==2.7.0
uvicorn==0.48.0
```

### Reproduction Steps

Library is unable to instantiate a tokenizer for any of the more recent models.

### Expected Behavior

Being able to instantiate a tokenizer all current models.

### Additional Context

We would like to (somewhat accurately) predict individual request costs beforehand to keep request latency in check.

### Suggested Solutions

It would be very helpful (= makes Mistral much more useful to our business) if MODEL_NAME_TO_TOKENIZER_CLS in tokens/tokenizers/mistral.py contains entries for the latest Mistral models.

juliendenize:
Hey #231 deprecates this method to inform which ones shoud be used instead.

PR #231, opened 2026-05-28
fix(tokenizer): point users at from_hf_hub on unknown model (#229)

NishchayMahor:
\`MistralTokenizer.from_model\` raises \`TokenizerException(\"Unrecognized model: X\")\` whenever the model name isn't in \`MODEL_NAME_TO_TOKENIZER_CLS\`. That mapping only covers older snapshots (latest entry is \`2411\`), so anything new — \`mistral-small-3\`, \`mistral-medium-3\`, \`magistral-medium\`, \`saba\`, ...  — hits this branch and the caller has no idea what to do next.

This swaps the bare message for one that points at the existing escape hatches and echoes the failing name so typos are obvious:

\`\`\`text
Unrecognized model: 'mistral-small-3-2501'. The static mapping in
MistralTokenizer.from_model only covers older model snapshots; newer or
custom models should be loaded with MistralTokenizer.from_hf_hub(repo_id=...)
or MistralTokenizer.from_file(...). Known names: codestral-2405, codestral-mamba-2407, ...
\`\`\`

I considered also adding new entries to \`MODEL_NAME_TO_TOKENIZER_CLS\` for the recent models, but I don't want to guess at the right tokenizer version (v3-tekken vs v7 vs v11 with image config) for each model — that decision really has to come from your side. Happy to send a follow-up once the mapping for the newer snapshots is published.

Refs #229.

## Tests

Existing \`tests/test_mistral_tokenizer.py::test_from_model\` already asserts \`TokenizerException\` for \`\"unknown-model\"\`. Extended it to also assert the message includes the offending name and the two escape hatches:

\`\`\`
$ pytest tests/test_mistral_tokenizer.py::TestMistralToknizer::test_from_model
1 passed in 0.40s
\`\`\`

### DIFF
--- a/src/mistral_common/tokens/tokenizers/mistral.py
+++ b/src/mistral_common/tokens/tokenizers/mistral.py
@@ -216,10 +216,7 @@ class MistralTokenizer(
 
     @classmethod
     def from_model(cls, model: str, strict: bool = True) -> "MistralTokenizer":
-        r"""Deprecated in favor of `from_hf_hub` or `from_file`, will be removed in 1.12.0.
-
-        The static `MODEL_NAME_TO_TOKENIZER_CLS` mapping is no longer kept in sync
-        with new model releases. Load tokenizers by repo or file instead.
+        r"""Deprecated in favor of `from_hf_hub` or `from_file`, will be removed in 1.13.0.
 
         Args:
             model: The model name.
@@ -229,7 +226,7 @@ class MistralTokenizer(
             The Mistral tokenizer for the given model.
         """
         warnings.warn(
-            "`MistralTokenizer.from_model` is deprecated and will be removed in 1.12.0. "
+            "`MistralTokenizer.from_model` is deprecated and will be removed in 1.13.0. "
             "Use `MistralTokenizer.from_hf_hub(...)` or `MistralTokenizer.from_file(...)` instead.",
             FutureWarning,
         )
--- a/src/mistral_common/tokens/tokenizers/mistral.py
+++ b/src/mistral_common/tokens/tokenizers/mistral.py
@@ -1,3 +1,4 @@
+import warnings
 from pathlib import Path
 from typing import Any, Callable, Generic
 
@@ -215,7 +216,10 @@ class MistralTokenizer(
 
     @classmethod
     def from_model(cls, model: str, strict: bool = True) -> "MistralTokenizer":
-        r"""Get the Mistral tokenizer for a given model.
+        r"""Deprecated in favor of `from_hf_hub` or `from_file`, will be removed in 1.12.0.
+
+        The static `MODEL_NAME_TO_TOKENIZER_CLS` mapping is no longer kept in sync
+        with new model releases. Load tokenizers by repo or file instead.
 
         Args:
             model: The model name.
@@ -224,17 +228,19 @@ class MistralTokenizer(
         Returns:
             The Mistral tokenizer for the given model.
         """
+        warnings.warn(
+            "`MistralTokenizer.from_model` is deprecated and will be removed in 1.12.0. "
+            "Use `MistralTokenizer.from_hf_hub(...)` or `MistralTokenizer.from_file(...)` instead.",
+            FutureWarning,
+        )
 
         if not strict:
             raise ValueError("strict has to be `True` since v1.10.0.")
 
         if model not in MODEL_NAME_TO_TOKENIZER_CLS:
-            known = ", ".join(sorted(MODEL_NAME_TO_TOKENIZER_CLS))
             raise TokenizerException(
-                f"Unrecognized model: {model!r}. The static mapping in MistralTokenizer.from_model "
-                f"only covers older model snapshots; newer or custom models should be loaded with "
-                f"MistralTokenizer.from_hf_hub(repo_id=...) or MistralTokenizer.from_file(...). "
-                f"Known names: {known}."
+                f"Unrecognized model: {model}. Use `MistralTokenizer.from_hf_hub(...)` "
+                f"or `MistralTokenizer.from_file(...)` to load newer or custom models."
             )
 
         return MODEL_NAME_TO_TOKENIZER_CLS[model]()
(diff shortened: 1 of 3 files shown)


src/mistral_common/tokens/tokenizers/mistral.py
```diff
@@ -229,7 +229,13 @@ def from_model(cls, model: str, strict: bool = True) -> "MistralTokenizer":
             raise ValueError("strict has to be `True` since v1.10.0.")
```
juliendenize, line 236: let's just deprecate here :)

src/mistral_common/tokens/tokenizers/mistral.py
```diff
@@ -229,7 +229,13 @@ def from_model(cls, model: str, strict: bool = True) -> "MistralTokenizer":
...
 
         if model not in MODEL_NAME_TO_TOKENIZER_CLS:
-            raise TokenizerException(f"Unrecognized model: {model}")
+            known = ", ".join(sorted(MODEL_NAME_TO_TOKENIZER_CLS))
+            raise TokenizerException(
+                f"Unrecognized model: {model!r}. The static mapping in MistralTokenizer.from_model "
```
juliendenize:
```suggestion
                f"Unrecognized model: {model}. The static mapping in MistralTokenizer.from_model "
```

src/mistral_common/tokens/tokenizers/mistral.py
```diff
@@ -229,7 +229,13 @@ def from_model(cls, model: str, strict: bool = True) -> "MistralTokenizer":
...
+                f"only covers older model snapshots; newer or custom models should be loaded with "
+                f"MistralTokenizer.from_hf_hub(repo_id=...) or MistralTokenizer.from_file(...). "
```
juliendenize: to be consistent
```suggestion
                f"MistralTokenizer.from_hf_hub(...) or MistralTokenizer.from_file(...). "
```

tests/test_mistral_tokenizer.py
```diff
@@ -26,9 +26,17 @@ def test_from_model(self) -> None:
             MistralTokenizer.from_model("open-mixtral-8x22b-2404").instruct_tokenizer, InstructTokenizerV3
         )
 
-        with pytest.raises(TokenizerException):
+        with pytest.raises(TokenizerException) as exc_info:
             MistralTokenizer.from_model("unknown-model")
 
+        # The error should point users at from_hf_hub / from_file so they can
+        # handle model snapshots that aren't in the static mapping, and echo
+        # the bad name so typos are obvious.
```
juliendenize:
```suggestion
```

juliendenize (changes requested):
Hey thanks for the contribution !

i think that the best would be to simply deprecate this from_model method and indeed point to from_file or from_hf_hub as we won't maintain the model mapping (it's been quite a while ^^)

raising as you did can be kept with minor change

lmk if you'd like to do the iteration otherwise i'll take over in another pr :)

NishchayMahor:
Really appreciate the thoughtful review @juliendenize :D. The deprecation framing is a much cleaner direction than my original "better error" attempt, and the consistency notes on the suggestion blocks made the diff way tighter. Genuinely learned something from this iteration, so thanks for taking the time on it!

Pushed the deprecation:
- `from_model` now emits a `FutureWarning` (modeled on `normalizer_for_tokenizer_version` in `normalize.py`) pointing at `from_hf_hub` / `from_file`, with `removed in 1.12.0` to match the existing precedent.
- Docstring updated to flag the deprecation up top.
- Applied all four of your inline suggestions: `{model}` instead of `{model!r}`, the `from_hf_hub(...)` / `from_file(...)` consistency, and dropped the extra comment in the test.
- Also dropped the `Known names: ...` listing from the error since you mentioned the map isn't being maintained, which felt misleading to enumerate going forward.
- Migrated the in-repo callers so test output stays quiet: `test_tokenizer_v7.py::test_from_model` is wrapped in `pytest.warns(FutureWarning)` (since that test exists to exercise the deprecated method), and the two helper calls in `test_mistral_tokenizer.py` now use `.v1()` / `.v3()`.

Local run is green (`13 passed`). Happy to iterate further if anything else needs adjusting. Really grateful for the steer either way :)

src/mistral_common/tokens/tokenizers/mistral.py
```diff
@@ -215,7 +216,10 @@ def v7(cls, is_mm: bool = False) -> "MistralTokenizer":
 
     @classmethod
     def from_model(cls, model: str, strict: bool = True) -> "MistralTokenizer":
-        r"""Get the Mistral tokenizer for a given model.
+        r"""Deprecated in favor of `from_hf_hub` or `from_file`, will be removed in 1.12.0.
```
juliendenize:
```suggestion
        r"""Deprecated in favor of `from_hf_hub` or `from_file`, will be removed in 1.13.0.
```

src/mistral_common/tokens/tokenizers/mistral.py
```diff
@@ -224,12 +228,20 @@ def from_model(cls, model: str, strict: bool = True) -> "MistralTokenizer":
         Returns:
             The Mistral tokenizer for the given model.
         """
+        warnings.warn(
+            "`MistralTokenizer.from_model` is deprecated and will be removed in 1.12.0. "
```
juliendenize:
```suggestion
            "`MistralTokenizer.from_model` is deprecated and will be removed in 1.13.0. "
```

src/mistral_common/tokens/tokenizers/mistral.py
```diff
@@ -215,7 +216,10 @@ def v7(cls, is_mm: bool = False) -> "MistralTokenizer":
 
     @classmethod
     def from_model(cls, model: str, strict: bool = True) -> "MistralTokenizer":
-        r"""Get the Mistral tokenizer for a given model.
+        r"""Deprecated in favor of `from_hf_hub` or `from_file`, will be removed in 1.12.0.
+
+        The static `MODEL_NAME_TO_TOKENIZER_CLS` mapping is no longer kept in sync
+        with new model releases. Load tokenizers by repo or file instead.
```
juliendenize:
```suggestion
```

juliendenize (commented):
Thanks for iterating looks good now

juliendenize (approved):

NishchayMahor:
Thanks @juliendenize :), means a lot to land this one. Appreciate the patient review and the steer toward deprecation over a louder error.

PR #233, opened 2026-06-03
Fix: forward continue_final_message in tekken normalizers (V7/V15)

### LINKED ISSUE #232: [BUG: Bug in the definition of continue_final_message that the Transforers library uses]
### Python -VV

```shell
Python 3.11.14 | packaged by conda-forge | (main, Oct 22 2025, 22:46:25) [GCC 14.3.0]
```

### Pip Freeze

```shell
datasets==4.8.4
docstring_parser==0.17.0
filelock==3.24.3
func_timeout==4.3.5
gguf==0.17.1
hf-xet==1.4.3
httpcore==1.0.9
httplib2==0.31.2
httptools==0.7.1
httpx==0.28.1
httpx-sse==0.4.3
huggingface_hub==1.9.2
interegular==0.3.3
jiter==0.13.0
jmespath==1.1.0
mistral_common==1.11.0
nltk==3.9.4
numpy==2.2.6
openai==2.24.0
openai-harmony==0.0.8
regex==2026.2.19
safetensors==0.7.0
scikit-learn==1.8.0
scipy==1.17.1
sentence-transformers==5.2.3
sentencepiece==0.2.1
tiktoken==0.12.0
tokenizers==0.22.2
torch==2.10.0
torch_c_dlpack_ext==0.1.5
torchaudio==2.10.0
torchvision==0.25.0
tqdm==4.67.3
transformers==5.5.3
tree-sitter==0.25.2
tree-sitter-python==0.25.0
triton==3.6.0
typer==0.24.1
uv==0.11.7
uvicorn==0.41.0
uvloop==0.22.1
vllm==0.19.1
zipp==3.23.0
```

### Reproduction Steps

```
from mistral_common.tokens.tokenizers.mistral import MistralTokenizer
from mistral_common.protocol.instruct.request import ChatCompletionRequest

tokenizer = MistralTokenizer.from_file("mistralai/Mistral-Small-3.2-24B-Instruct-2506")

request = ChatCompletionRequest.from_openai(
    messages=[
        {"role": "user", 'content': 'test'},
        {"role": "assistant", 'content': 'test'},
    ],
    continue_final_message=True,
)
print(tokenizer.encode_chat_completion(request))
```
This code produce the following output: `tokens=[1, 3, 4417, 4, 4417, 2] text='<s>[INST]test[/INST]test</s>' prefix_ids=[4417, 2] images=[] audios=[]`
I believe this is uncorrect as we have the token 2 (`</s>`) which correspond to the EOS token in Mistral-Small-3.2-24B-Instruct-2506. This token should not appear since in [mistral_common/tokens/tokenizers/instruct.py #L1145](https://github.com/mistralai/mistral-common/blob/main/src/mistral_common/tokens/tokenizers/instruct.py#L1145C9-L1146C54) we have:
```
        if not message.prefix and not continue_message:
            curr_tokens.append(self.tokenizer.eos_id)
```
which idicate that the continue_message option should not include the EOS token.

This code is inspired by the use of transformers for Mistral's tokenizer ([tokenization_mistral_common.py#L1178](https://github.com/huggingface/transformers/blob/main/src/transformers/tokenization_mistral_common.py#L1178)) and this implementation is used by vLLM so this would produce unappropriate answer from the model when launched in a vLLM server.

### Expected Behavior

The previous code snippet should result in `tokens=[1, 3, 4417, 4, 4417] text='<s>[INST]test[/INST]test' prefix_ids=[4417, 2] images=[] audios=[]`.

### Additional Context

This bug was described in the Transformers library in issue [#46326](https://github.com/huggingface/transformers/issues/46326) but I believe it should be corrected in `mistral-common`.
I originally encouter it when using vLLM to serve a Mistral's model on a server but the behavior of the model was deeply impact by the option `continue_final_message=True`.
As things stand, it is impossible to use a Mistral model with vLLM and to write manually the begin of the assistant answer. This can be a disadvantage in certain use cases.

### Suggested Solutions

_No response_

matdou:
## What

`continue_final_message` is dropped on the tekken path. `InstructRequestNormalizerV7.from_chat_completion_request` and `InstructRequestNormalizerV15.from_chat_completion_request` build the `InstructRequest` without forwarding the flag, so it defaults to `False`. `continue_message` is then `False` in `encode_instruct`, and EOS is appended at `instruct.py` L1145 even when `continue_final_message=True`.

The base `InstructRequestNormalizer.from_chat_completion_request` forwards it correctly, so SentencePiece tokenizers (v1–v3) are unaffected; only tekken tokenizers (v7/v11/v13/v15) hit this — including Mistral-Small-3.2-24B.

## Repro

```python
from mistral_common.tokens.tokenizers.mistral import MistralTokenizer
from mistral_common.protocol.instruct.request import ChatCompletionRequest

tok = MistralTokenizer.v7(is_mm=True)
req = ChatCompletionRequest.from_openai(
    messages=[{"role": "user", "content": "a"}, {"role": "assistant", "content": "b"}],
    continue_final_message=True,
)
tok.encode_chat_completion(req).tokens
# before: [1, 3, 1032, 4, 1055, 2]   <- trailing EOS (2)
# after:  [1, 3, 1032, 4, 1055]
```

## Fix

Forward `continue_final_message=request.continue_final_message` in both overrides.

## Test

Adds a regression test through the full `encode_chat_completion` path. The existing `test_tokenize_assistant_message_continue_final_message` only exercises `encode_instruct` on a hand-built `InstructRequest`, which bypasses the normalizer — that's why this regressed silently.

Fixes #232. Also resolves the downstream report at huggingface/transformers#46326; no transformers-side change is needed once this lands.

### DIFF
--- a/tests/test_tokenizer_v7.py
+++ b/tests/test_tokenizer_v7.py
@@ -518,3 +518,22 @@ def test_assistant_tool_call_and_content(request: pytest.FixtureRequest, tekkeni
     tokens_2 = mistral_tokenizer.encode_chat_completion(chat_completion_request)
 
     assert tokens == tokens_2.tokens
+
+
+def test_encode_chat_completion_continue_final_message() -> None:
+    # Regression: the tekken normalizers (V7/V15) dropped continue_final_message
+    # when converting ChatCompletionRequest -> InstructRequest, so the full
+    # encode_chat_completion path appended EOS. encode_instruct was unaffected,
+    # which is why the existing test missed it. See #232.
+    tokenizer = MistralTokenizer.v7(is_mm=True)
+    eos_id = tokenizer.instruct_tokenizer.tokenizer.eos_id
+
+    request: ChatCompletionRequest = ChatCompletionRequest(
+        messages=[UserMessage(content="a"), AssistantMessage(content="b")],
+        continue_final_message=True,
+    )
+    encoded = tokenizer.encode_chat_completion(request)
+
+    assert encoded.tokens == [1, 3, 1032, 4, 1055]
+    assert encoded.tokens[-1] != eos_id
+    assert eos_id not in encoded.prefix_ids
--- a/tests/test_tokenizer_v7.py
+++ b/tests/test_tokenizer_v7.py
@@ -521,10 +521,6 @@ def test_assistant_tool_call_and_content(request: pytest.FixtureRequest, tekkeni
 
 
 def test_encode_chat_completion_continue_final_message() -> None:
-    # Regression: the tekken normalizers (V7/V15) dropped continue_final_message
-    # when converting ChatCompletionRequest -> InstructRequest, so the full
-    # encode_chat_completion path appended EOS. encode_instruct was unaffected,
-    # which is why the existing test missed it. See #232.
     tokenizer = MistralTokenizer.v7(is_mm=True)
     eos_id = tokenizer.instruct_tokenizer.tokenizer.eos_id
 
--- a/tests/test_tokenizer_v7.py
+++ b/tests/test_tokenizer_v7.py
@@ -518,3 +518,18 @@ def test_assistant_tool_call_and_content(request: pytest.FixtureRequest, tekkeni
     tokens_2 = mistral_tokenizer.encode_chat_completion(chat_completion_request)
 
     assert tokens == tokens_2.tokens
+
+
+def test_encode_chat_completion_continue_final_message() -> None:
+    tokenizer = MistralTokenizer.v7(is_mm=True)
+    eos_id = tokenizer.instruct_tokenizer.tokenizer.eos_id
+
+    request: ChatCompletionRequest = ChatCompletionRequest(
+        messages=[UserMessage(content="a"), AssistantMessage(content="b")],
+        continue_final_message=True,
+    )
+    encoded = tokenizer.encode_chat_completion(request)
+
+    assert encoded.tokens == [1, 3, 1032, 4, 1055]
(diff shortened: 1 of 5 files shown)


juliendenize:
Hey thanks for the contribution, i created another pr #234 because i cannot edit yours and i added a bunch of tests. However if you prefer you could add the changes so that we can merge yours :)

matdou:
> Hey thanks for the contribution, i created another pr #234 because i cannot edit yours and i added a bunch of tests. However if you prefer you could add the changes so that we can merge yours :)

Hey, thanks for the extra tests! I've pulled them into this PR, so we should be good to merge this one and close #234.

juliendenize (approved):
Thanks for iterating LGTM :)

tests/test_tokenizer_v7.py
```diff
@@ -518,3 +518,22 @@ def test_assistant_tool_call_and_content(request: pytest.FixtureRequest, tekkeni
     tokens_2 = mistral_tokenizer.encode_chat_completion(chat_completion_request)
 
     assert tokens == tokens_2.tokens
+
+
+def test_encode_chat_completion_continue_final_message() -> None:
+    # Regression: the tekken normalizers (V7/V15) dropped continue_final_message
+    # when converting ChatCompletionRequest -> InstructRequest, so the full
+    # encode_chat_completion path appended EOS. encode_instruct was unaffected,
+    # which is why the existing test missed it. See #232.
```
juliendenize:
```suggestion
```

PR #255, opened 2026-06-26
Raise ValueError instead of ValidationError in audio chunk validators

sarathfrancis90:
Constructing an `AudioChunk` with empty input raises a confusing `TypeError` instead of a validation error:

```python
>>> from mistral_common.protocol.instruct.chunk import AudioChunk
>>> AudioChunk(input_audio="")
TypeError: ValidationError.__new__() missing 1 required positional argument: 'line_errors'
```

The `should_not_be_empty` validator raises `pydantic.ValidationError("...")` directly, but in Pydantic v2 that type has no public string constructor, so the validator itself blows up before it can report the real problem. The deprecated `RawAudio.format` validator has the same issue.

I switched both to a plain `ValueError`, which is what every other validator in the library already does and which Pydantic wraps into a proper `ValidationError` carrying the message. Empty `str`, whitespace-only `str`, and empty `bytes` now all raise a clear validation error.

Added a parametrized test in `tests/test_audio.py` for those three inputs. Ran the full test suite, ruff, ruff format, mypy and the doctests locally, all green.

### DIFF
--- a/src/mistral_common/protocol/instruct/chunk.py
+++ b/src/mistral_common/protocol/instruct/chunk.py
@@ -6,7 +6,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal
 from urllib.parse import urlparse
 
-from pydantic import ConfigDict, Field, ValidationError, field_validator, model_validator
+from pydantic import ConfigDict, Field, field_validator, model_validator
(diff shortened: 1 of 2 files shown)


juliendenize (approved):
makes sense thanks

PR #256, opened 2026-06-27
Decode normal tokens under SpecialTokenPolicy.RAISE in SentencePiece

sarathfrancis90:
While decoding with `SpecialTokenPolicy.RAISE` on a SentencePiece tokenizer, I noticed it returns the raw SentencePiece pieces instead of decoded text when the tokens contain no special token:

```python
tok = MistralTokenizer.v3().instruct_tokenizer.tokenizer
ids = tok.encode("Hello world, how are you?", bos=False, eos=False)
tok.decode(ids, SpecialTokenPolicy.IGNORE)  # 'Hello world, how are you?'
tok.decode(ids, SpecialTokenPolicy.RAISE)   # '▁Hello▁world,▁how▁are▁you?'  <- pieces, not text
```

RAISE was sharing the KEEP code path (`_decode_with_special_tokens`), which keeps non-special tokens as pieces. That's documented behavior for KEEP, but RAISE is only meant to raise when a special token is present and otherwise decode normally — the way IGNORE here and Tekken already do. So a perfectly valid "decode and make sure there are no special tokens" call comes back garbled.

I split RAISE out of that path: raise if any token is special, otherwise decode through SentencePiece like IGNORE. KEEP is left as-is. Added tests for both the no-special-token decode and the raise-on-special-token case; ruff, mypy, and the full pytest + doctest suites pass.

### DIFF
--- a/src/mistral_common/tokens/tokenizers/sentencepiece.py
+++ b/src/mistral_common/tokens/tokenizers/sentencepiece.py
@@ -218,8 +218,13 @@ class SentencePieceTokenizer(Tokenizer):
                 f"Expected `special_token_policy` to be a SpecialTokenPolicy, got {type(special_token_policy)}."
             )
 
-        if special_token_policy in [SpecialTokenPolicy.KEEP, SpecialTokenPolicy.RAISE]:
-            return self._decode_with_special_tokens(tokens, special_token_policy)
+        if special_token_policy == SpecialTokenPolicy.RAISE:
+            if any(self.is_special(tok) for tok in tokens):
+                raise ValueError("Decoding `tokens` that contain special tokens with special_token_policy=RAISE.")
+            return self._model.decode(tokens)  # type: ignore[no-any-return]
+
+        if special_token_policy == SpecialTokenPolicy.KEEP:
+            return self._decode_with_special_tokens(tokens)
 
         return self._model.decode(tokens)  # type: ignore[no-any-return]
 
@@ -227,13 +232,13 @@ class SentencePieceTokenizer(Tokenizer):
         r"""Convert the given token id to a token piece."""
         return self._model.id_to_piece(token_id)  # type: ignore
 
-    def _decode_with_special_tokens(self, tokens: list[int], special_token_policy: SpecialTokenPolicy) -> str:
+    def _decode_with_special_tokens(self, tokens: list[int]) -> str:
+        # KEEP policy: special tokens are rendered as their piece string and normal tokens are kept
+        # as SentencePiece pieces (see the note in `decode`).
         text_list = []
         curr_tokens: list[int] = []
         for tok in tokens:
             if self.is_special(tok):
-                if special_token_policy == SpecialTokenPolicy.RAISE:
-                    raise ValueError("Decoding `tokens` that contain special tokens with special_token_policy=RAISE.")
                 if curr_tokens:
                     text_list.extend([self.id_to_piece(tok) for tok in curr_tokens])
                     curr_tokens = []
--- a/src/mistral_common/tokens/tokenizers/sentencepiece.py
+++ b/src/mistral_common/tokens/tokenizers/sentencepiece.py
@@ -220,7 +220,7 @@ class SentencePieceTokenizer(Tokenizer):
 
         if special_token_policy == SpecialTokenPolicy.IGNORE:
             decoded = self._model.decode(tokens)
-            assert isinstance(decoded, str), f"Mode decoded a {type(decoded)}, not a string."
+            assert isinstance(decoded, str), f"Sentencepiece model decoded a {type(decoded)}, not a string."
             return decoded
 
         return self._decode_with_special_tokens(tokens, special_token_policy)
--- a/src/mistral_common/tokens/tokenizers/sentencepiece.py
+++ b/src/mistral_common/tokens/tokenizers/sentencepiece.py
@@ -218,29 +218,27 @@ class SentencePieceTokenizer(Tokenizer):
                 f"Expected `special_token_policy` to be a SpecialTokenPolicy, got {type(special_token_policy)}."
             )
 
-        if special_token_policy == SpecialTokenPolicy.RAISE:
-            if any(self.is_special(tok) for tok in tokens):
-                raise ValueError("Decoding `tokens` that contain special tokens with special_token_policy=RAISE.")
+        if special_token_policy == SpecialTokenPolicy.IGNORE:
             return self._model.decode(tokens)  # type: ignore[no-any-return]
 
-        if special_token_policy == SpecialTokenPolicy.KEEP:
-            return self._decode_with_special_tokens(tokens)
-
-        return self._model.decode(tokens)  # type: ignore[no-any-return]
+        return self._decode_with_special_tokens(tokens, special_token_policy)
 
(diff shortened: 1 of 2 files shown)


src/mistral_common/tokens/tokenizers/sentencepiece.py
```diff
@@ -218,22 +218,27 @@ def decode(self, tokens: list[int], special_token_policy: SpecialTokenPolicy = S
                 f"Expected `special_token_policy` to be a SpecialTokenPolicy, got {type(special_token_policy)}."
             )
 
-        if special_token_policy in [SpecialTokenPolicy.KEEP, SpecialTokenPolicy.RAISE]:
-            return self._decode_with_special_tokens(tokens, special_token_policy)
+        if special_token_policy == SpecialTokenPolicy.RAISE:
+            if any(self.is_special(tok) for tok in tokens):
```
juliendenize: this makes another loop on the input which can be costly for long sequence.

src/mistral_common/tokens/tokenizers/sentencepiece.py
```diff
@@ -218,22 +218,27 @@ def decode(self, tokens: list[int], special_token_policy: SpecialTokenPolicy = S
...
         return self._model.id_to_piece(token_id)  # type: ignore
 
-    def _decode_with_special_tokens(self, tokens: list[int], special_token_policy: SpecialTokenPolicy) -> str:
+    def _decode_with_special_tokens(self, tokens: list[int]) -> str:
+        # KEEP policy: special tokens are rendered as their piece string and normal tokens are kept
+        # as SentencePiece pieces (see the note in `decode`).
         text_list = []
         curr_tokens: list[int] = []
         for tok in tokens:
             if self.is_special(tok):
```
juliendenize, line 236: raise here

src/mistral_common/tokens/tokenizers/sentencepiece.py
```diff
@@ -218,22 +218,27 @@ def decode(self, tokens: list[int], special_token_policy: SpecialTokenPolicy = S
...
+            return self._decode_with_special_tokens(tokens)
 
         return self._model.decode(tokens)  # type: ignore[no-any-return]
 
     def id_to_piece(self, token_id: int) -> str:
         r"""Convert the given token id to a token piece."""
         return self._model.id_to_piece(token_id)  # type: ignore
 
-    def _decode_with_special_tokens(self, tokens: list[int], special_token_policy: SpecialTokenPolicy) -> str:
+    def _decode_with_special_tokens(self, tokens: list[int]) -> str:
```
juliendenize: adding policy here might make more sense to decode and raise if needed

juliendenize (changes requested):
Thanks for the contribution, i think the end result makes a lot of sense.

However i believe it would actually be better to put the check for special inside the private method to raise there instead of making two loops.

I added comment for clarity :)

sarathfrancis90:
Good call — folded the RAISE check into `_decode_with_special_tokens` so it raises on the first special token in a single pass and dropped the separate `any()` scan over the input. When a RAISE sequence has no special tokens, the trailing run is decoded via `_model.decode` so normal text still comes back correctly (not pieces); KEEP is unchanged.

src/mistral_common/tokens/tokenizers/sentencepiece.py
```diff
@@ -218,24 +218,27 @@ def decode(self, tokens: list[int], special_token_policy: SpecialTokenPolicy = S
...
+        return self._decode_with_special_tokens(tokens, special_token_policy)
 
     def id_to_piece(self, token_id: int) -> str:
         r"""Convert the given token id to a token piece."""
         return self._model.id_to_piece(token_id)  # type: ignore
 
     def _decode_with_special_tokens(self, tokens: list[int], special_token_policy: SpecialTokenPolicy) -> str:
+        # KEEP renders special tokens (and the normal tokens around them) as their SentencePiece
+        # pieces; RAISE rejects any special token in a single pass and otherwise decodes the
+        # (all-normal) sequence normally.
```
juliendenize:
```suggestion
```

src/mistral_common/tokens/tokenizers/sentencepiece.py
```diff
@@ -244,7 +247,11 @@ def _decode_with_special_tokens(self, tokens: list[int], special_token_policy: S
                 curr_tokens.append(tok)
 
         if curr_tokens:
-            text_list.extend([self.id_to_piece(tok) for tok in curr_tokens])
+            if special_token_policy == SpecialTokenPolicy.RAISE:
+                # Reached only when no special token was present, so decode the whole run.
```
juliendenize:
```suggestion
```

tests/test_sentencepiece.py
```diff
@@ -48,3 +49,19 @@ def test_sentencepiece_tokenizer_num_specials_property(tokenizer_v7: SentencePie
 
     # Test that num_special_tokens matches the length of special_ids
     assert num_special == len(tokenizer_v7.special_ids)
+
+
+def test_decode_raise_policy_decodes_normal_tokens(tokenizer_v7: SentencePieceTokenizer) -> None:
+    # With no special tokens present, RAISE should decode normal tokens to text just like IGNORE,
+    # not return raw SentencePiece pieces (which is the documented behavior of KEEP only).
```
juliendenize:
```suggestion
```

juliendenize (approved):
Thanks looks good now ! just going to remove comments

src/mistral_common/tokens/tokenizers/sentencepiece.py
```diff
@@ -218,10 +218,10 @@ def decode(self, tokens: list[int], special_token_policy: SpecialTokenPolicy = S
                 f"Expected `special_token_policy` to be a SpecialTokenPolicy, got {type(special_token_policy)}."
             )
 
-        if special_token_policy in [SpecialTokenPolicy.KEEP, SpecialTokenPolicy.RAISE]:
-            return self._decode_with_special_tokens(tokens, special_token_policy)
+        if special_token_policy == SpecialTokenPolicy.IGNORE:
+            return self._model.decode(tokens)  # type: ignore[no-any-return]
```
juliendenize:
```suggestion
            decoded = self._model.decode(tokens)
            assert isinstance(decoded, str), f"Mode decoded a {type(decoded)}, not a string."
            return decoded
```

src/mistral_common/tokens/tokenizers/sentencepiece.py
```diff
@@ -218,10 +218,12 @@ def decode(self, tokens: list[int], special_token_policy: SpecialTokenPolicy = S
                 f"Expected `special_token_policy` to be a SpecialTokenPolicy, got {type(special_token_policy)}."
             )
 
-        if special_token_policy in [SpecialTokenPolicy.KEEP, SpecialTokenPolicy.RAISE]:
-            return self._decode_with_special_tokens(tokens, special_token_policy)
+        if special_token_policy == SpecialTokenPolicy.IGNORE:
+            decoded = self._model.decode(tokens)
+            assert isinstance(decoded, str), f"Mode decoded a {type(decoded)}, not a string."
```
juliendenize:
```suggestion
            assert isinstance(decoded, str), f"Sentencepiece model decoded a {type(decoded)}, not a string."
```

PR #257, opened 2026-06-30
Avoid SentencePiece decode type ignore

juliendenize:
## Summary
- Replace the SentencePiece `decode` return `type: ignore[no-any-return]` with a small `_model_decode` helper.
- Use `isinstance(decoded, str)` to narrow the `Any` return from `sentencepiece` before returning.
- Reuse the helper for the new `RAISE` all-normal-token decode path from PR #256.

## Verification
- `python -m mypy src/mistral_common/tokens/tokenizers/sentencepiece.py tests/test_sentencepiece.py`
- `python -m pytest tests/test_sentencepiece.py`
- `python -m ruff check src/mistral_common/tokens/tokenizers/sentencepiece.py tests/test_sentencepiece.py`

## Note
This is a follow-up to #256. I could not push directly to the contributor fork branch from this environment, so this branch contains #256 plus the type-ignore cleanup.

### DIFF
--- a/src/mistral_common/tokens/tokenizers/sentencepiece.py
+++ b/src/mistral_common/tokens/tokenizers/sentencepiece.py
@@ -218,14 +218,20 @@ class SentencePieceTokenizer(Tokenizer):
                 f"Expected `special_token_policy` to be a SpecialTokenPolicy, got {type(special_token_policy)}."
             )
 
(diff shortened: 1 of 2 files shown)


juliendenize:
Closing in favor of leaving the suggested change directly on #256.

PR #260, opened 2026-07-08
Auto-select grammar variant from tokenizer

juliendenize:
## Summary
Removes the caller-driven `reasoning: bool` flag from grammar generation. `GrammarFactory.select_jinja_template()` now auto-selects the variant from the tokenizer:

- **think** grammar (`[THINK]`/`[/THINK]` special tokens) when version >= v13 AND both think tokens are present
- **base** grammar otherwise

Base already accepts a literal `<think>...</think>` block as free text, so no template change was needed for that.

## Changes
- `select_jinja_template()` takes no arguments; selection is derived from tokenizer version + `[THINK]`/`[/THINK]` presence.
- Fixes a latent bug where a v13+ tokenizer missing `[THINK]` rendered a broken grammar with `None` tokens — now falls back to base.
- `_GrammarVariant.plain_think` and its template file are retained but no longer reachable via the factory.
- Guidance tests updated: v11 cases retargeted to base semantics; added coverage for the v13-without-think fallback and base-accepts-literal-`<think>`.

Scope limited to the guidance module. Protocol/tokenizer/chat-template reasoning code is untouched.

## Breaking change
`GrammarFactory.select_jinja_template(reasoning=...)` no longer accepts the `reasoning` argument.

## Verification
- ruff check + ruff format: clean
- mypy `src/mistral_common/guidance/`: clean
- `pytest tests/guidance/`: 339 passed

### DIFF
--- a/src/mistral_common/guidance/grammar_factory.py
+++ b/src/mistral_common/guidance/grammar_factory.py
@@ -35,14 +35,11 @@ def _validate_mode_and_tools(mode: ToolChoice, tools: list[Tool] | None) -> None
 
 
 @lru_cache()
-def _cached_get_jinja_template(tokenizer_version: TokenizerVersion, reasoning: bool) -> str:
-    if not reasoning:
-        jinja_key = _GrammarVariant.base
-    elif tokenizer_version < TokenizerVersion.v13:
-        jinja_key = _GrammarVariant.plain_think
-    else:
+def _cached_get_jinja_template(tokenizer_version: TokenizerVersion, has_think_tokens: bool) -> str:
+    if tokenizer_version >= TokenizerVersion.v13 and has_think_tokens:
         jinja_key = _GrammarVariant.think
-
+    else:
+        jinja_key = _GrammarVariant.base
     return JINJA_PATHS[jinja_key].read_text(encoding="utf-8")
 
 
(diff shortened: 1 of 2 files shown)


src/mistral_common/guidance/grammar_factory.py
```diff
@@ -206,16 +203,22 @@ def _get_optional_special_token_lark(self, token_name: str) -> str | None:
...
+          `[THINK]` and `[/THINK]` special tokens are registered.
+        - Returns the `base` template in all other cases.
 
         Returns:
             The jinja template content as a string.
         """
-        return _cached_get_jinja_template(tokenizer_version=self._tokenizer.version, reasoning=reasoning)
+        has_think_tokens = (
+            SpecialTokens.begin_think.value in self._special_token_map
+            and SpecialTokens.end_think.value in self._special_token_map
```
joa-stdn: not a big fan because this hard-codes special tokens here as well, but not super opposed if there is no other solution
I'm not sure why we need to know whether the tokenizer has think token, is it just that the grammar will not compile if it does not, and we already released the tekken json files without?

juliendenize: yes grammar cannot compile without it and yes also for the second question hence this heurisitic. I think if you find it okish it would be nice to keep as is :)

joa-stdn (approved):

PR #262, opened 2026-07-12
Fix seed field not bridged in SpeechRequest OpenAI conversion

winklemad:
No linked issue — found by reading the code.

## The bug

`SpeechRequest` uses the field name `random_seed`, while the OpenAI request format uses `seed`. Neither converter bridges the two:

- `to_openai()` returns the raw `model_dump`, so it emits `random_seed` (a mistral-common-internal name) instead of `seed`.
- `from_openai()` keeps only keys that are model fields, so an OpenAI request's `seed` is never mapped to `random_seed` and is silently dropped:

```python
from mistral_common.protocol.speech.request import SpeechRequest

SpeechRequest.from_openai({"input": "hi", "voice": "female", "seed": 123}).random_seed
# None   -> the seed provided in the OpenAI request is lost

SpeechRequest(input="hi", voice="female", random_seed=123).to_openai()["seed"]
# KeyError: 'seed'   -> the key is emitted as "random_seed" instead
```

## Smoking gun

The sibling `TranscriptionRequest` already bridges the two names in both directions:

```python
# TranscriptionRequest.to_openai
openai_request["seed"] = openai_request.pop("random_seed")
# TranscriptionRequest.from_openai
seed = openai_request.get("seed")
...
converted_dict["random_seed"] = seed
```

`SpeechRequest` does neither, so it is inconsistent with its sibling.

## Fix

Translate `random_seed` ⇄ `seed` in both `SpeechRequest.to_openai` and `SpeechRequest.from_openai`, mirroring `TranscriptionRequest`. A seed now survives the round trip and an OpenAI-style request with `seed` is honored.

## Testing

Added `test_convert_speech_request_seed`, asserting `to_openai()` emits `seed` (not `random_seed`) and `from_openai({"seed": ...})` sets `random_seed`. It fails before the change (`KeyError: 'seed'`) and passes after; the full `tests/test_converters.py` suite passes (104 passed). `ruff check` / `ruff format --check` are clean.

### DIFF
--- a/src/mistral_common/protocol/speech/request.py
+++ b/src/mistral_common/protocol/speech/request.py
@@ -54,7 +54,6 @@ class SpeechRequest(BaseCompletionRequest):
 
             openai_request["ref_audio"] = buffer
 
-        # OpenAI uses "seed"; mistral-common uses "random_seed".
         openai_request["seed"] = openai_request.pop("random_seed")
         openai_request.update(kwargs)
 
@@ -95,7 +94,6 @@ class SpeechRequest(BaseCompletionRequest):
         if isinstance(voice, dict):
             converted_dict["voice"] = voice["id"]
 
-        # OpenAI uses "seed"; mistral-common uses "random_seed".
         converted_dict["random_seed"] = seed
 
         return cls(**converted_dict)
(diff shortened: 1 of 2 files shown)


juliendenize (approved):
Thanks for the contribution, made a minor update and will merge :)

PR #242, opened 2026-06-14
Support Responses API

Halcyonhal9:
Accept Responses API input_text/output_text/input_image content types

## What

Change `_convert_openai_content_chunks` to also accept the OpenAI **Responses API**
content types `input_text`, `output_text` (→ `TextChunk`) and `input_image`
(→ `ImageURLChunk`).

## Why

`ChunkTypes` mirrors the OpenAI **Chat Completions** content types (`text`,
`image_url`, ...). 

The [Responses API](https://platform.openai.com/docs/api-reference/responses)
names text and image content differently — `input_text`/`output_text` for text
and `input_image` for images. This means parsing structured Responses content fails
with e.g. `'input_text' is not a valid ChunkTypes` /
`'input_image' is not a valid ChunkTypes`.

**This breaks the Responses API for Mistral models served on vLLM.** Serving a
Mistral model on **vLLM 0.23.0** (with `mistral_common` 1.11.3) using
`--tokenizer-mode mistral` and exposing `/v1/responses` hands request content
straight to `mistral_common`. A client sending the standard Responses input
shape — `{"type": "input_text", "text": "..."}` or
`{"type": "input_image", "image_url": "..."}`, which the official OpenAI client
libraries emit — gets an HTTP **400**, making `/v1/responses` unusable for text
and image input alike.

## The proposed change

Map the Responses content-type spellings onto their Chat Completions
equivalents in `_convert_openai_content_chunks` via a small alias table
(`input_text`/`output_text` → `text`, `input_image` → `image_url`). The chunk's
`type` field is rewritten because the discriminated chunk models require the
canonical `type`; `input_image`'s `image_url` is a bare string, which
`ImageURLChunk` already accepts.

## Tests / checks

- Regression tests for text (`input_text`/`output_text`) and image
  (`input_image`) content.
- `pytest tests/test_converters.py` → 116 passed.
- `ruff check`, `ruff format --check`, `mypy` → all clean.

A patched version of mistral_common 1.11.3 was tested on vLLM 0.23.0, using hf model mistralai/Mistral-Medium-3.5-128B and the responses api.

PR #263, opened 2026-07-12
Accept OpenAI Responses input content types

Halcyonhal9:
### Summary

Make `from_openai` (via `_convert_openai_content_chunks`) accept the OpenAI **Responses API** input content-type spellings and map them onto their Chat Completions equivalents:

- `input_text` / `output_text` → `ChunkTypes.text`
- `input_image` → `ChunkTypes.image_url` (Responses carries `image_url` as a bare string, which `ImageURLChunk` already accepts)

The change is small and additive — it normalizes the discriminator `type` before validation, so structured Responses input is tokenized instead of raising `ValueError: 'input_text' is not a valid ChunkTypes`. No new request type is introduced; existing Chat Completions behaviour is unchanged. Unit tests are included.

> Recreates #242, which was closed due to inactivity (apologies, I missed the question in the comments section). The branch has been resynced to the latest `main` and has no conflicts.

### Why the Responses API is required here (not Chat Completions)

The earlier review asked why we don't just use `ChatCompletionRequest`. The reason is **reasoning support**, and it is grounded in the official OpenAI specification rather than a preference:

- The upstream application consumes model **reasoning output**, and it uses an AI harness that implements **the official OpenAI API specification** (no vendor-specific extensions).
- In the official spec, **reasoning output is exclusive to the Responses API**. A Responses result contains a first-class `reasoning` output item (with `summary[]` / `content[]`), and reasoning is requested via `reasoning.effort`.
- The **Chat Completions** specification has **no field for reasoning content** in the response. The only reasoning-related value it officially returns is a *token count* (`usage.completion_tokens_details.reasoning_tokens`) — the reasoning text itself is never returned. (The `reasoning` / `reasoning_content` fields some servers add to Chat Completions messages are **non-standard vendor extensions**, which a spec-only harness does not consume.)

Because the harness only speaks the official spec, it can obtain reasoning **only** over the Responses API. When those Responses-shaped requests are served by a Mistral model whose tokenization goes through `mistral_common` (e.g. vLLM's `--tokenizer-mode mistral`), the request body uses `input_text` / `output_text` / `input_image`, and `mistral_common` currently rejects it with `400 'input_text' is not a valid ChunkTypes` — even though the underlying content is exactly what the Chat Completions chunks already represent.

This PR removes that mismatch at the parsing boundary, so Responses input can be tokenized without every caller having to pre-rewrite content types.

### Supporting references (official OpenAI docs)

- **Reasoning guide** — reasoning is accessed and returned through the Responses API; "Reasoning models work better with the Responses API": https://platform.openai.com/docs/guides/reasoning
- **Responses API — object** (reasoning output item with `summary` / `content`): https://platform.openai.com/docs/api-reference/responses/object
- **Responses API — create** (`reasoning.effort` request field; `input_text` / `input_image` input parts): https://platform.openai.com/docs/api-reference/responses/create
- **Chat Completions — object** (response `message` has no reasoning field; reasoning appears only as a `reasoning_tokens` count under `usage`): https://platform.openai.com/docs/api-reference/chat/object

### Scope / compatibility

- Additive and backward-compatible: canonical Chat Completions types are untouched; only the three Responses spellings are newly accepted and mapped to existing chunk types.
- No new public API, no new dependency.
- Tests cover `input_text`, `output_text`, and `input_image`.

### DIFF
--- a/src/mistral_common/protocol/instruct/chunk.py
+++ b/src/mistral_common/protocol/instruct/chunk.py
@@ -456,11 +456,32 @@ UserContentChunk = Annotated[
 ]
 
 
+# The OpenAI Responses API spells text/image content types differently from the
+# Chat Completions API that ``ChunkTypes`` mirrors. Map the Responses spellings
+# onto their Chat Completions equivalents so structured Responses input/output
+# is accepted instead of raising e.g. "'input_text' is not a valid ChunkTypes".
+# (``input_image`` carries ``image_url`` as a bare string, which
+# ``ImageURLChunk`` already accepts.)
+_RESPONSES_CONTENT_TYPE_ALIASES: dict[str, ChunkTypes] = {
+    "input_text": ChunkTypes.text,
+    "output_text": ChunkTypes.text,
+    "input_image": ChunkTypes.image_url,
+}
+
+
 def _convert_openai_content_chunks(openai_content_chunks: dict[str, Any]) -> ContentChunk:
     content_type_str = openai_content_chunks.get("type")
 
     if content_type_str is None:
         raise ValueError("Content chunk must have a type field.")
+
+    # Normalize OpenAI Responses content types to the Chat Completions spellings.
+    # The discriminated chunk models require the canonical ``type``, so rewrite
+    # the field too.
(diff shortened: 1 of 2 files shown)


juliendenize:
Hi thanks for the contrib , we don't plan to support Responses API for now

PR #265, opened 2026-07-17
Add MLX (Apple Silicon) inference example to the docs

Lee-Gihun:
Adds an MLX (Apple Silicon) inference example to `docs/examples/inference.md`: encode with mistral-common, generate at token level with mlx-lm, decode with the same tokenizer. No nav change.

### Context
With transformers v5, `AutoTokenizer` loads Mistral models that ship a `tekken.json` as `MistralCommonBackend`. Wrappers that assume fast-tokenizer attributes currently break on this (ml-explore/mlx-lm#1576), and there is no guidance on the mistral-common side for MLX users. The token-level pattern documented here is the path that works today and keeps prompt construction on the official tokenizer.

Example validated verbatim with mistral-common 1.11.5, mlx-lm 0.31.3, and `mlx-community/Ministral-3-3B-Instruct-2512-4bit`.

### DIFF
--- a/docs/examples/inference.md
+++ b/docs/examples/inference.md
@@ -9,6 +9,7 @@ We have a few examples of how to use the library with our models:
   - [Audio](#audio)
 - [Fill-in-the-middle (FIM) Completion](#fim)
 - [Audio Transcription](#audio-transcription)
+- [MLX (Apple Silicon)](#mlx-apple-silicon)
 
 ## Chat Completion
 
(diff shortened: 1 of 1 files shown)


juliendenize:
Hey this snippet is related to how to load using Transformers which is not the scope of this repo focused on processing. Therefore i'm closing this one.

Lee-Gihun:
Makes sense, thanks for the review. Agreed this is about loading and inference rather than processing.

PR #269, opened 2026-07-20
Fix v15 tool-result rendering when thinking is enabled

Th0rgal:
## Summary

- pass `support_thinking=false` when rendering simple v15 tool results if the generated `render_content` macro declares the thinking parameter
- update the v15 thinking golden templates
- add regression coverage for text-only and image-capable v15 templates

## Problem

When thinking support is enabled, `render_content` declares `support_thinking` as a required macro parameter. The simple tool-result branch did not pass it:

```jinja
{{- render_content(message['content'], 'tool message contents', supported_types_desc='text and image', support_images=true) -}}
```

Strict Jinja rendering therefore accepts the first assistant tool call but fails as soon as the client appends a `role: tool` result:

```text
Not enough arguments provided to macro 'render_content'
```

The official Leanstral 1.5 template currently contains the affected generated call:
https://huggingface.co/mistralai/Leanstral-1.5-119B-A6B/blob/81592da95d94ab0439bfce16df1d55b402e598b6/chat_template.jinja#L221

This change renders tool content with `support_thinking=false`, matching the intended tool-result content types and the explicit handling used by other message-role branches.

## Reproduction

A minimal failing conversation is:

1. user message
2. assistant message with native `tool_calls`
3. client-appended `role: tool` result

The new test renders that sequence for v15 with thinking enabled, both with and without image support. Before this patch, both cases fail at Jinja macro evaluation; after it, they render `[TOOL_RESULTS]ok[/TOOL_RESULTS]`.

An independent llama.cpp deployment using the generated fixed template was also exercised end to end:

- native structured tool calls: 10/10
- textual pseudo-calls: 0/10
- direct tool-result continuations: 3/3 HTTP 200
- routed tool-result continuations: 3/3 HTTP 200
- stock Mistral Vibe completed multi-turn tool execution

Sanitized reproduction and evidence:
https://github.com/Th0rgal/mistral-vibe-leanstral-repro/blob/main/COMPARISON.md

## Validation

- `pytest tests/integrations/chat_templates/unit/test_v15.py -q` — 27 passed
- `pytest tests/integrations/chat_templates/unit -q` — 180 passed
- Ruff check and format check — passed
- mypy on the changed Python files — passed
- llama.cpp `test-chat-template --with-tools` on the generated fixed template — passed

### DIFF
--- a/src/mistral_common/integrations/chat_templates/template_generator.py
+++ b/src/mistral_common/integrations/chat_templates/template_generator.py
@@ -1522,6 +1522,8 @@ def _generate_tool_message_handling(config: TemplateConfig) -> str:
                 if config.audio_support:
                     desc_parts.append("audio")
                 tool_rc_args += f", supported_types_desc='{_join_types_desc(desc_parts)}'"
+            if config.any_thinking_support:
+                tool_rc_args += ", support_thinking=false"
             if config.image_support:
                 tool_rc_args += ", support_images=true"
             if config.audio_support:
--- a/tests/data/chat_templates/v15_image_think.jinja
+++ b/tests/data/chat_templates/v15_image_think.jinja
@@ -218,7 +218,7 @@
     {#- Tool messages (multimodal). #}
     {%- elif message['role'] == 'tool' %}
         {{- '[TOOL_RESULTS]' -}}
-        {{- render_content(message['content'], 'tool message contents', supported_types_desc='text and image', support_images=true) -}}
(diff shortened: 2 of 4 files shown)


juliendenize:
Hey thanks for the PR, i've created another one that should meet your goal but more suited to mistral-common features https://github.com/mistralai/mistral-common/pull/271

juliendenize:
Actually tool should not be allowed to think, i'll close this one and make it explicit in the codebase, again ty for your contribution and let us know if it brings an inconvenience, but allowing this would mean our models would be OOD

PR #271, opened 2026-07-20
Render v15 tool-result thinking correctly and pass all render_content args explicitly

juliendenize:
## Summary

Fixes v15 tool-result rendering when thinking is enabled, and eliminates all implicit `render_content` macro arguments in the generated chat templates.

> **Supersedes #269** (same topic). #269 stops the crash by passing `support_thinking=false` for tool results, but that *rejects* thinking chunks in tool messages — which mistral-common actually supports and encodes. This PR instead renders tool-result thinking (`support_thinking=true`), matching the tokenizer, and additionally makes every `render_content` call pass all declared arguments explicitly. #269 can be closed in favor of this PR.

## Problem

When thinking support is enabled, the generated `render_content` macro declares `support_thinking` (and `supported_types_desc`) as required parameters, but the v15 simple tool-result branch did not pass them:

```jinja
{{- render_content(message['content'], 'tool message contents', supported_types_desc='text and image', support_images=true) -}}
```

Strict Jinja rendering then fails as soon as a `role: tool` message is appended (`parameter 'supported_types_desc' was not provided` / not enough macro arguments).

## Why `support_thinking=true` (not `false`)

mistral-common v15 tool messages support thinking content:
- `ToolMessage._allowed_content_chunks` includes `ThinkChunk`.
- `MistralRequestValidatorV15._validate_tool_content_chunks` accepts all chunk types.
- `InstructTokenizerV13.encode_tool_message` (inherited by V15) encodes list content via `_encode_content_chunks`, so a `ThinkChunk` becomes `[THINK]...[/THINK]`.

Empirically the tokenizer produces `...[TOOL_RESULTS][THINK]tool reasoning here[/THINK]Sunny, 25C[/TOOL_RESULTS]...`. The chat template must render this identically, so it needs `support_thinking=true` with `thinking` in `supported_types_desc`. Passing `false` (as in #269) would make the template reject content the tokenizer accepts.

## Changes

- **Fix the v15 tool-result branch** to pass `support_thinking=true` and include `thinking` in `supported_types_desc`.
- **Centralize call construction** in a new `_render_content_call(...)` helper so every `render_content` call emits *every* argument the macro declares for the config — no reliance on Jinja implicit `undefined`. This also fixed pre-existing implicit-arg cases in `v3_image`, `v3_image_spm`, and `v7_image_spm` (missing `supported_types_desc` / `support_images` / `initial_prev_img`).
- **All arguments are named**, including `content` and `context_name`: calls now read `render_content(content=..., context_name=..., ...)`.
- **Regenerated all affected golden templates** accordingly.
- **Tests:** focused v15 unit tests for tool-result thinking rendering/acceptance, plus a `REQUEST_TOOL_THINKING_TRAIN` conversation fixture wired into `_get_conversations`, so tool-with-reasoning is covered by both the template-vs-template parity (`test_parity.py`) and the mistral-common-vs-chat-template parity (`transformers/test_core_parity.py`).

## Validation

- `pytest tests/integrations/chat_templates -q` — all pass (incl. transformers cross-implementation parity with the new tool-reasoning fixture).
- Golden parity (`test_dynamic_matches_static`) and rendered-output parity (`test_dynamic_template_produces_same_output`) pass for every config — rendered output for pre-existing configs is unchanged (only made explicit).
- `ruff check`, `ruff format --check`, and `mypy` on changed files — pass.

### DIFF
--- a/tests/integrations/chat_templates/unit/test_v15.py
+++ b/tests/integrations/chat_templates/unit/test_v15.py
@@ -4,7 +4,12 @@ import pytest
 
 from mistral_common.integrations.chat_templates.chat_templates import generate_chat_template
 from mistral_common.tokens.tokenizers.base import TokenizerVersion
-from tests.integrations.chat_templates.helpers import TestConfig, _load_golden_template, _make_config, render_template
+from tests.integrations.chat_templates.helpers import (
+    TestConfig,
+    _load_golden_template,
+    _make_config,
+    render_template,
+)
 
 
 class TestV15ModelSettings:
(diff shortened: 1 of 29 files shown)


tests/integrations/chat_templates/unit/test_v15.py
```diff
@@ -471,3 +476,135 @@ def test_pre_v15_image_template_rejects_image_in_tool(self) -> None:
...
+            {"role": "user", "content": "Hello"},
+            {
+                "role": "assistant",
+                "content": "",
+                "tool_calls": [{"id": "abcdefghi", "function": {"name": "f", "arguments": "{}"}}],
+            },
+            {
+                "role": "tool",
+                "content": [
+                    {"type": "thinking", "thinking": "secret reasoning"},
```
joa-stdn, line 503: I don't understand this, thinking is not allowed in tool response, if it's allowed it's a bug IMO

juliendenize: making another pr !

PR #272, opened 2026-07-20
Add auto-detecting chat template generation from tokenizer file

juliendenize:
## Summary

Add `convert_tokenizer_to_chat_template(tokenizer_file, system_prompt, use_special_token_variables)` that auto-detects a tokenizer's capabilities from its file and generates the matching HuggingFace chat template, removing the need to pass 8 explicit flags to `generate_chat_template`.

Auto-detection loads the tokenizer via `MistralTokenizer.from_file` and derives:
- **version** from the tokenizer
- **spm** (SentencePiece vs Tekken backend)
- **image support** (`image_encoder` present)
- **audio support** (`audio_encoder` present)
- **thinking support** (`[THINK]`/`[/THINK]` special tokens, i.e. `BEGIN_THINK` set on v13+)
- **plain thinking** for v11 (`<think>` text tags), guarded off when audio is present

## Changes

- New `convert_tokenizer_to_chat_template` in `src/mistral_common/integrations/chat_templates/chat_templates.py`, delegating to the existing `generate_chat_template`.
- CLI `scripts/generate_chat_template.py` gains an auto-detect mode via `--tokenizer_file`. It is mutually exclusive with the manual capability flags (`--version`, `--spm`, `--image`, `--audio`, `--thinking`, `--plain_thinking`) and errors clearly if combined; `--version` is now required only in manual mode.

## Test Plan

- [x] `uv run pytest tests/integrations/chat_templates/unit/test_convert_tokenizer.py tests/test_generate_chat_template_cli.py` (28 passed)
- [x] `uv run ruff check .` and `uv run mypy` on changed files clean
- [x] Tests cover tekken v3 image, v7 audio, v11 plain-think, v11+audio guard, v13 thinking, spm v7, system-prompt embedding, special-token-variables toggle, unrecognized-file error, and CLI conflict/missing-version paths

### DIFF
--- a/src/mistral_common/integrations/chat_templates/chat_templates.py
+++ b/src/mistral_common/integrations/chat_templates/chat_templates.py
@@ -6,7 +6,7 @@ from mistral_common.integrations.chat_templates.template_generator import (
 from mistral_common.integrations.chat_templates.template_generator import (
     build_chat_template as _build_chat_template,
 )
-from mistral_common.tokens.tokenizers.base import TokenizerVersion
+from mistral_common.tokens.tokenizers.base import SpecialTokens, TokenizerVersion
 from mistral_common.tokens.tokenizers.mistral import MistralTokenizer
 from mistral_common.tokens.tokenizers.sentencepiece import SentencePieceTokenizer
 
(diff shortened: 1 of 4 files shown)


src/mistral_common/integrations/chat_templates/chat_templates.py
```diff
@@ -60,3 +64,59 @@ def generate_chat_template(
...
+    """
+    mistral_tokenizer = MistralTokenizer.from_file(tokenizer_file)
+    instruct_tokenizer = mistral_tokenizer.instruct_tokenizer
+    tokenizer = instruct_tokenizer.tokenizer
+
+    version = mistral_tokenizer.version
+    spm = isinstance(tokenizer, SentencePieceTokenizer)
+    image_support = instruct_tokenizer.image_encoder is not None
+    audio_support = instruct_tokenizer.audio_encoder is not None
+    thinking_support = getattr(instruct_tokenizer, "BEGIN_THINK", None) is not None
```
joa-stdn: I don't really like this tbh, this is one of the difficulties with having just tekken.json available IMO
But even if we choose this, I think the way is more to use `tokenizer.get_special_token(SpecialTokens.begin_think)` and check that this is the same as `end_think`? 
A bit like this https://github.com/mistralai/mistral-common/blob/main/src/mistral_common/guidance/grammar_factory.py#L231 but ideally we don't want to use private variables?

joa-stdn (approved):

src/mistral_common/integrations/chat_templates/chat_templates.py
juliendenize: switched to get_special thanks

PR #273, opened 2026-07-21
Validate special_token_policy strings when decoding

sarathfrancis90:
## What

`Tekkenizer.decode` and `SentencePieceTokenizer.decode` accept `special_token_policy` as either a `SpecialTokenPolicy` or a plain string, but they only guard the *type* and never validate the *value*. An invalid string slips straight through the guard and is then compared against the enum members, matching none of them.

The result is a silent footgun: the same typo behaves differently on each tokenizer and neither raises.

```python
tekken.decode(ids, "kepe")  # meant "keep" -> special tokens silently DROPPED (like IGNORE)
spm.decode(ids, "kepe")     # meant "keep" -> special tokens silently KEPT  (like KEEP)
```

So a caller who mistypes the policy gets wrong output back with no indication anything went wrong, and the behaviour is inconsistent between the two tokenizers.

## Fix

Route the policy through the `SpecialTokenPolicy(...)` constructor after the type guard, so an invalid value raises the same clear `ValueError` already used everywhere else (and covered by `test_tokenizer_base.py`). Valid strings and enum members keep working exactly as before.

## Testing

- Added tests for both tokenizers: a valid policy string decodes identically to the enum, and an invalid string now raises.
- `uv run pytest tests/ --ignore=tests/integrations`, doctests, `ruff check`, `ruff format --check`, `mypy .`, and `pre-commit run --all-files` all pass.

### DIFF
--- a/src/mistral_common/tokens/tokenizers/sentencepiece.py
+++ b/src/mistral_common/tokens/tokenizers/sentencepiece.py
@@ -213,11 +213,13 @@ class SentencePieceTokenizer(Tokenizer):
         Returns:
             The decoded string.
         """
-        if not isinstance(special_token_policy, (str, SpecialTokenPolicy)):
+        try:
+            special_token_policy = SpecialTokenPolicy(special_token_policy)
+        except ValueError as e:
+            valid = ", ".join(repr(p.value) for p in SpecialTokenPolicy)
             raise ValueError(
-                f"Expected `special_token_policy` to be a SpecialTokenPolicy, got {type(special_token_policy)}."
-            )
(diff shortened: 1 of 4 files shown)


juliendenize:
Hey thanks for the contribution ! Nice catch, I updated it a bit because i think we can now just use SpecialTokenPolicy constructor to catch the valid values and avoid multiple checks

juliendenize (approved):

sarathfrancis90:
Makes sense — funneling it through the `SpecialTokenPolicy` constructor keeps the valid values in one place instead of my separate checks. Thanks for tidying it up and for the quick merge.

PR #274, opened 2026-07-22
Emit all render_content macro args explicitly

juliendenize:
## What

Refactors the chat-template generator so every `render_content(...)` macro call passes all its arguments explicitly as keywords, via a single `_render_content_call(...)` helper. Replaces the ad-hoc per-site argument string-building in the system/user/assistant/tool handlers.

## Why

Call sites previously relied on Jinja's implicit `undefined` for omitted macro args, and each site rebuilt the argument string by hand. Centralizing this in one helper makes the emitted arguments explicit, ordered, and consistent, and removes duplication.

## Behavior

Pure refactor — rendered template output is unchanged vs `main`. Only the generated `.jinja` text changes (args become keyword form / explicit), so the golden fixtures under `tests/data/chat_templates/` are regenerated to match.

Note: the tool-message call passes `support_thinking=false` (and does not add `thinking` to its supported types), preserving `main`'s behavior — tool messages continue to reject thinking chunks at the template level (the macro raises).

## Tests

- Unit tests for `_render_content_call` (minimal/full/tool-ban/ordering/spm cases), including exact-string assertions locking argument order and formatting.
- New test asserting a v15 (image+think) template raises `ValueError` when a tool message dict carries a thinking chunk, and does not leak the thinking text into the error.
- Regenerated goldens verified by the existing parity suite (`test_dynamic_matches_static`, plus rendered-output parity tests).
- `pytest tests/integrations/chat_templates`, `ruff check`, `ruff format --check`, `mypy` all pass.

### DIFF
--- a/src/mistral_common/integrations/chat_templates/template_generator.py
+++ b/src/mistral_common/integrations/chat_templates/template_generator.py
@@ -241,6 +241,85 @@ def _join_types_desc(parts: list[str]) -> str:
     return ", ".join(parts[:-1]) + " and " + parts[-1]
 
 
+def _render_content_call(
+    config: TemplateConfig,
+    content_expr: str,
+    context_name: str,
+    *,
+    supported_types_desc: str,
+    support_thinking: bool,
+    support_images: bool,
+    support_audio: bool,
+    initial_prev_img: str = "false",
+) -> str:
+    r"""Build a `render_content(...)` Jinja call string with all args emitted explicitly.
+
+    Every argument that the `render_content` macro declares for `config` is included
+    as a keyword argument in macro-declared order, so no call site relies on Jinja
+    implicit `undefined`. Arguments for features absent from `config` are omitted
+    because the macro does not declare them.
+
+    Args:
+        config: Template configuration controlling which macro parameters exist.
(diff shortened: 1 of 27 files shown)


src/mistral_common/integrations/chat_templates/template_generator.py
```diff
@@ -891,25 +970,24 @@ def _generate_system_message_handling(config: TemplateConfig) -> str:
...
+    elif config.system_supports_audio:
+        sys_types_desc = "text and audio"
+    else:
+        sys_types_desc = "text"
+
+    rc = _render_content_call(
+        config=config,
+        content_expr="message['content']",
+        context_name="system message contents",
+        supported_types_desc=sys_types_desc,
```
joa-stdn, line 984: can't this arg be determined from the others?

joa-stdn (approved):

src/mistral_common/integrations/chat_templates/template_generator.py
juliendenize: it's the same helper for audio which cannot solely be infered by others

PR #282, opened 2026-07-30
Deprecate Tokenized.text in favor of tokenizer.decode

juliendenize:
## What

Deprecates `Tokenized.text` (now a property backed by a private `_text` cache) in favor of `tokenizer.decode(tokens=tokenized.tokens, special_token_policy=SpecialTokenPolicy.KEEP)`, migrates every internal call site off the removed field, and adds regression coverage plus a shared test helper.

## Why

`text` was a plain field on `Tokenized` that got eagerly computed and stored on every encode call, even when callers never touched it. Making it a deprecated property (still eagerly decoded into a private cache at construction, exactly like the old field) lets us warn callers on access during the deprecation window without changing behavior; once the property is fully removed in 1.13.0, the eager decode cost goes away too.

An earlier version of this change kept a live back-reference to the tokenizer instead and decoded lazily, to also avoid the eager-decode cost during the deprecation window. That added real complexity for no real payoff: a custom `__eq__` (to keep equality from comparing tokenizer instances), a `__getstate__` override (to keep the tokenizer out of pickled payloads), and four call sites that had to remember to wire up the backref (three didn't, silently turning `None` into a raised exception on `encode_speech_request`). Caching the decoded string instead removes all of that: strings pickle trivially, equality works out of the box, and every construction site just needs one line, matching what it already did before this deprecation.

## Behavior

- `Tokenized.text` still returns the same decoded string it always did, but now warns once (`DeprecationWarning`) per process, and returns `None` for a `Tokenized` where the decoded text was never cached (matching the original field's default and every call site's prior behavior, including the ones that never set `text` on `main`).
- `Tokenized` equality is plain pydantic default equality (no custom `__eq__`): it now also compares the cached `_text`, which is fine since it's cheap and deterministic for a given `tokens`.

## Tests

- `tests/test_tokenizer_base.py`: coverage for the deprecated `text` property (decodes correctly, returns `None` when not cached, warns exactly once, participates in equality) and that pickling round-trips `_text` correctly (unlike the tokenizer-backref approach, nothing needs to be stripped).
- `tests/utils.py`: new `decode_keep(tokenizer, tokenized)` helper collapsing ~40 repeated `tokenizer.decode(tokens=..., special_token_policy=SpecialTokenPolicy.KEEP)` call sites across the suite.
- Every existing test/doc call site that used `tokenized.text`/`encoded.text` migrated to the new helper (one equality assert also updated since `Tokenized(..., text=...)` is no longer a valid constructor call).
- Full suite (`1425 passed, 17 skipped`), `ruff check`, `ruff format --check`, `mypy`, and doctests all pass.

## Notes

`tests/utils.py` is a plain helper module (not `test_utils.py`, so pytest won't collect it as tests). Heads up: the in-progress `test-refactor-pr2-instruct` branch documents a planned `tests.utils.regenerate_registry` *package* — if that branch lands first, `tests/utils.py` (a file) will need to be reconciled with `tests/utils/` (a directory) at merge time.

### DIFF
--- a/src/mistral_common/tokens/tokenizers/base.py
+++ b/src/mistral_common/tokens/tokenizers/base.py
@@ -212,35 +212,17 @@ class Tokenized(MistralBase):
     prefix_ids: list[int] | None = None
     images: list[np.ndarray] = Field(default_factory=list)
     audios: list[Audio] = Field(default_factory=list)
-    _tokenizer: "InstructTokenizer | None" = PrivateAttr(default=None)
-
-    def __eq__(self, other: object) -> bool:
-        r"""Compares only the public fields, matching pydantic's default equality.
-
-        The private `_tokenizer` back-reference used by the deprecated `text` property is not
-        part of the value and must not affect equality.
-        """
-        if other.__class__ is not self.__class__:
-            return NotImplemented
-        return self.__dict__ == other.__dict__
-
-    def __getstate__(self) -> dict[str, Any]:
-        r"""Strips `_tokenizer` so pickling a `Tokenized` never serializes the tokenizer that produced it."""
-        state = super().__getstate__()
-        private = state.get("__pydantic_private__")
-        if private:
-            state["__pydantic_private__"] = {**private, "_tokenizer": None}
-        return state
+    _text: str | None = PrivateAttr(default=None)
 
     @property
     def text(self) -> str | None:
-        r"""Decodes the token ids into a string using the tokenizer that produced them.
+        r"""The text representation of the tokens, decoded with special tokens kept.
 
         Deprecated: Use `tokenizer.decode(tokens=tokenized.tokens, special_token_policy=SpecialTokenPolicy.KEEP)`
         instead. Will be removed in 1.13.0.
 
         Returns:
-            The decoded string, or `None` if this `Tokenized` was not produced by a tokenizer.
+            The decoded string, or `None` if it was not set by the tokenizer that produced this `Tokenized`.
         """
         warn_once(
             key="Tokenized.text",
@@ -251,9 +233,7 @@ class Tokenized(MistralBase):
             category=DeprecationWarning,
             stacklevel=2,
         )
-        if self._tokenizer is None:
-            return None
-        return self._tokenizer.decode(tokens=self.tokens, special_token_policy=SpecialTokenPolicy.KEEP)
+        return self._text
 
 
 class Tokenizer(ABC):
--- a/src/mistral_common/tokens/tokenizers/base.py
+++ b/src/mistral_common/tokens/tokenizers/base.py
@@ -5,9 +5,10 @@ from pathlib import Path
 from typing import Any, Generic, TypeVar
 
 import numpy as np
-from pydantic import ConfigDict, Field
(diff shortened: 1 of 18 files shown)


src/mistral_common/tokens/tokenizers/base.py
```diff
@@ -198,20 +199,61 @@ class Tokenized(MistralBase):
...
+        >>> tokenized = Tokenized(tokens=[1, 2, 3], prefix_ids=[1], images=[], audios=[])
     """
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
     tokens: list[int]
-    text: str | None = None
     prefix_ids: list[int] | None = None
     images: list[np.ndarray] = Field(default_factory=list)
     audios: list[Audio] = Field(default_factory=list)
+    _tokenizer: "InstructTokenizer | None" = PrivateAttr(default=None)
```
joa-stdn: so this is only for the purpose of the migration? not familiar enough with OSS on tokenized.text to know if this is classical (I find storing this a bit weird), wouldn't it be equivalent to raise warning right now and then remove it later?

juliendenize: I really don't like storing this either but the issue of having text kept is that i would raise deprecated warning regardless on whether the end user specifically asks for .text.<br><br>Usually when you deprecate a feature, you want the end user to adopt the correct workflow and then stop warning about stuff.

juliendenize: based on our discussion i made _text instead of _tokenizer :)

src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -225,13 +225,15 @@ def encode_instruct(
...
-        return Tokenized(
+        tokenized = Tokenized(
             tokens=tokens,
-            text=self.decode(tokens, special_token_policy=SpecialTokenPolicy.KEEP),
             prefix_ids=prefix_ids,
             images=images,
             audios=audios,
         )
+        # TODO: remove once 1.13.0 lands.
+        tokenized._text = self.decode(tokens=tokens, special_token_policy=SpecialTokenPolicy.KEEP)
```
joa-stdn, line 235: Yeah sounds good, we can also say
```
Tokenized(
...,   
_text=self.decode(tokens=tokens, special_token_policy=SpecialTokenPolicy.KEEP)
)
```
but both are fine I guess?

joa-stdn (approved):

src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -1092,10 +1102,12 @@ def _encode_streaming_transcription(self, request: TranscriptionRequest) -> Toke
 
         return Tokenized(
             tokens=tokens,
-            text=self.decode(tokens, special_token_policy=SpecialTokenPolicy.KEEP),
             audios=audios,
+            # TODO: remove once 1.13.0 lands.
+            _text=self.decode(tokens=tokens, special_token_policy=SpecialTokenPolicy.KEEP),
         )
 
+
```
juliendenize:
```suggestion
```

PR #284, opened 2026-08-01
Set reference audio filename for speech multipart uploads

Mr-Neutr0n:
## Summary

- Preserve the detected format of reference audio passed through `SpeechRequest.to_openai()`.
- Name multipart streams `audio.<format>` for both base64 and raw-byte inputs.

## Root cause

`SpeechRequest.to_openai()` returned an unnamed `BytesIO`, so multipart clients defaulted to filename `upload` and `Content-Type: application/octet-stream`. The fix mirrors `TranscriptionRequest` by detecting or preserving the audio format and setting the filename to `audio.<format>`.

## Tests

- 4 focused cases: WAV and FLAC, each as base64 and raw bytes
- 1,145 unit tests passed; 52 doctests passed; 454 integration tests passed
- Full Ruff check, Ruff format check, and mypy check passed

AI assistance was used during investigation and implementation; the behavior, diff, and test results were independently verified before submission.

### DIFF
--- a/src/mistral_common/protocol/speech/request.py
+++ b/src/mistral_common/protocol/speech/request.py
@@ -55,6 +55,7 @@ class SpeechRequest(BaseCompletionRequest):
                 sf.write(buffer, audio.audio_array, audio.sampling_rate, format=audio.format)
                 buffer.seek(0)
 
+            # OpenAI's client uses the filename extension from .name to set the Content-Type.
             buffer.name = f"audio.{fmt}"
             openai_request["ref_audio"] = buffer
 
--- a/src/mistral_common/protocol/speech/request.py
+++ b/src/mistral_common/protocol/speech/request.py
@@ -3,6 +3,7 @@ from typing import Any
 
 from mistral_common.imports import assert_soundfile_installed, is_soundfile_installed
 from mistral_common.protocol.base import BaseCompletionRequest
+from mistral_common.protocol.instruct.chunk import _detect_audio_format
 from mistral_common.tokens.tokenizers.audio import Audio
 
 if is_soundfile_installed():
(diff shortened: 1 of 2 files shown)


src/mistral_common/protocol/speech/request.py
```diff
@@ -45,13 +46,16 @@ def to_openai(self, **kwargs: Any) -> dict[str, Any]:
         if self.ref_audio is not None:
             if isinstance(self.ref_audio, bytes):
                 buffer = io.BytesIO(self.ref_audio)
+                fmt = _detect_audio_format(self.ref_audio)
```
juliendenize, line 49: This introduces a new failure mode. Before this PR raw bytes were never inspected; now `sf.info()` runs and raises on anything it can't parse:

```
SpeechRequest(input="hi", ref_audio=b"not audio").to_openai()
→ ValueError: Failed to detect audio format. Verify that the given file is valid wav or mp3.
```

The error path is untested. `test_convert_transcription_bytes_invalid_format` already exists for the transcription side; a speech equivalent would cover the branch this PR makes reachable.

src/mistral_common/protocol/speech/request.py
```diff
@@ -45,13 +46,16 @@ def to_openai(self, **kwargs: Any) -> dict[str, Any]:
...
             else:
                 audio = Audio.from_base64(self.ref_audio)
+                fmt = audio.format.lower()
 
                 buffer = io.BytesIO()
                 sf.write(buffer, audio.audio_array, audio.sampling_rate, format=audio.format)
                 buffer.seek(0)
 
+            buffer.name = f"audio.{fmt}"
```
juliendenize, line 59: Please carry over the explanatory comment from `transcription/request.py`:

```python
# OpenAI's client uses the filename extension from .name to set the Content-Type.
```

tests/test_converters.py
```diff
@@ -1248,6 +1248,30 @@ def test_convert_speech_request_from_openai() -> None:
...
 
+@pytest.mark.parametrize("fmt", ["wav", "flac"])
+def test_speech_to_openai_base64_ref_audio_filename(fmt: str) -> None:
+    audio = _make_fake_audio(0.5)
+    request = SpeechRequest(input="Hello world", ref_audio=audio.to_base64(fmt))
+
+    buffer = request.to_openai()["ref_audio"]
+
+    assert isinstance(buffer, io.BytesIO)
+    assert buffer.name == f"audio.{fmt}"
```
juliendenize, line 1259: The transcription counterpart also round-trips the buffer:

```python
recovered = Audio.from_bytes(buffer.getvalue())
assert np.allclose(recovered.audio_array, audio.audio_array, atol=1e-3)
```

Worth adding here too !

juliendenize (commented):
Hey thanks for your contribution, I left some comments to make it mergeable, lmk if you need help!

Mr-Neutr0n:
Addressed the review: added the invalid raw-bytes regression, documented why the buffer filename controls multipart content type, and verified the encoded audio round-trip. https://github.com/Mr-Neutr0n/mistral-common/commit/93e00e47ca6d1fe71fedba1d92fb219c0c95d846

juliendenize (approved):
Thanks for iterating

PR #289, opened 2026-08-19
Keep at least one pixel per axis when resizing an image

arthi-arumugam-git:
## The bug

An image with a very extreme aspect ratio crashes the encoder on valid input. At the production config (`image_patch_size=16`, `max_image_size=1024`), a 4 by 10000 pixel image raises a bare `AssertionError`:

```python
cfg = ImageConfig(image_patch_size=16, max_image_size=1024, spatial_merge_size=1)
enc = ImageEncoder(cfg, SpecialImageIDs(img=10, img_break=11, img_end=12))
enc(ImageChunk(image=Image.new("RGB", (4, 10000))))   # AssertionError
```

Same for 1 by 4096 and 3 by 8000. Nothing about those images is invalid: they are just tall.

## Why

In `_image_to_num_tokens`, when the image is larger than `max_image_size` the two sides are divided by the ratio and rounded. For a ratio above twice the shorter side, `round()` takes that side to 0, so the token count for that axis is `(0 - 1) // patch + 1`, which is 0, and `__call__` then trips its own `assert w > 0`.

## The change

Both axes clamp to a minimum of one pixel after the resize, so an extremely thin image encodes as a single row or column of tokens instead of crashing. Nothing else changes: any image whose shorter side survives the resize is unaffected, and the existing expectations in `test_image_to_num_tokens` still hold.

## Tests

`test_image_to_num_tokens_extreme_aspect_ratio` covers 1x512, 4x10000, 10000x4 and 2x4096 at both `spatial_merge_size` values, asserting both that the axis counts are at least 1 and that the encoder produces the expected token count. It fails on `main` at both parameter values and passes here. The rest of `tests/test_image.py` is green (22 passed).

### DIFF
--- a/tests/test_image.py
+++ b/tests/test_image.py
@@ -42,21 +42,20 @@ def test_image_to_num_tokens(spatial_merge_size: int, special_token_ids: Special
 
 
 @pytest.mark.parametrize("spatial_merge_size", [1, 2])
-def test_image_to_num_tokens_extreme_aspect_ratio(spatial_merge_size: int, special_token_ids: SpecialImageIDs) -> None:
-    # For a very wide or very tall image the resize used to round the shorter side
-    # down to 0 pixels, which made the token count for that axis 0 and crashed the
-    # encoder on valid input. The shorter side must clamp to at least one token.
+@pytest.mark.parametrize("size", [(1, 512), (4, 10000), (10000, 4), (2, 4096)])
+def test_image_to_num_tokens_extreme_aspect_ratio(
+    special_token_ids: SpecialImageIDs, size: tuple[int, int], spatial_merge_size: int
+) -> None:
     image_config = ImageConfig(
         image_patch_size=16 // spatial_merge_size, max_image_size=128, spatial_merge_size=spatial_merge_size
     )
     image_encoder = ImageEncoder(image_config, special_token_ids)
 
-    for size in [(1, 512), (4, 10000), (10000, 4), (2, 4096)]:
-        img = Image.new("RGB", size, "red")
-        w_tokens, h_tokens = image_encoder._image_to_num_tokens(img)
-        assert w_tokens >= 1 and h_tokens >= 1, (size, w_tokens, h_tokens)
-        encoding = image_encoder(ImageChunk(image=img))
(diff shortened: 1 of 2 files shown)


tests/test_image.py
```diff
@@ -41,6 +41,24 @@ def test_image_to_num_tokens(spatial_merge_size: int, special_token_ids: Special
         assert image_encoder._image_to_num_tokens(img) == (exp1, exp2)
 
 
+@pytest.mark.parametrize("spatial_merge_size", [1, 2])
+def test_image_to_num_tokens_extreme_aspect_ratio(spatial_merge_size: int, special_token_ids: SpecialImageIDs) -> None:
+    # For a very wide or very tall image the resize used to round the shorter side
+    # down to 0 pixels, which made the token count for that axis 0 and crashed the
+    # encoder on valid input. The shorter side must clamp to at least one token.
```
juliendenize: can be removed imo :)

```suggestion
```

tests/test_image.py
```diff
@@ -41,6 +41,24 @@ def test_image_to_num_tokens(spatial_merge_size: int, special_token_ids: Special
...
+    image_config = ImageConfig(
+        image_patch_size=16 // spatial_merge_size, max_image_size=128, spatial_merge_size=spatial_merge_size
+    )
+    image_encoder = ImageEncoder(image_config, special_token_ids)
+
+    for size in [(1, 512), (4, 10000), (10000, 4), (2, 4096)]:
```
juliendenize: could you use parametrized instead ?

juliendenize (commented):
Thanks for the contribution !

Looks good but could you make parametrized tests and clean up a bit the comments before i approve ?

arthi-arumugam-git:
Both done, thanks.

The four sizes are parametrized now, stacked with the existing `spatial_merge_size` one, so it runs as 8 cases rather than one test with a loop inside it. That also means a failing size shows up in the test name instead of only in an assert message, so I dropped that message too.

Comment block in the test is gone, and I cut the one in `image.py` down to a single line.

All 8 cases still fail on main and pass here, and `tests/test_image.py` is 28 passed.

juliendenize (approved):
Awesome, thanks for iterating :)

PR #292, opened 2026-08-23
Reject AudioConfig where audio_length_per_tok floors to zero

shoemoney:
#253 added a guard requiring `sampling_rate >= frame_rate` to prevent a ZeroDivisionError in `num_audio_tokens`. That guard does not cover the case where the ratio is positive but smaller than `hop_length`: `audio_length_per_tok` is `int((sampling_rate / frame_rate) / hop_length)`, which floors to 0 and the same ZeroDivisionError survives.

One-line repro on current main:

```python
AudioConfig(sampling_rate=200, frame_rate=2.0, encoding_config=AudioSpectrogramConfig(num_mel_bins=80, hop_length=160, window_size=400)).num_audio_tokens(500)  # ZeroDivisionError
```

This adds a second assert in `__post_init__`, in the same style as #253, requiring `audio_length_per_tok > 0` with a message naming `sampling_rate`, `frame_rate`, and `hop_length`, plus a test alongside #253's. Full suite: 1434 passed, 17 skipped (baseline 1433 plus the new test).

### DIFF
--- a/src/mistral_common/tokens/tokenizers/audio.py
+++ b/src/mistral_common/tokens/tokenizers/audio.py
@@ -352,6 +352,11 @@ class AudioConfig:
             f"sampling_rate must be >= frame_rate so that each token spans at least one sample, "
             f"got sampling_rate={self.sampling_rate} and frame_rate={self.frame_rate}"
         )
+        assert self.audio_length_per_tok > 0, (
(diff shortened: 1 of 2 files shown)


juliendenize (approved):
Makes sense, thanks for your contribution !

PR #293, opened 2026-08-24
Reject null tool call IDs before encoding

juliendenize:
## Summary
- reject null tool call IDs during validation for tokenizer v13 and newer
- preserve v11 support with a version-specific validator
- cover all validation modes and version dispatch

## Testing
- `PYTHONPATH=src /Users/julien.denize/workspace/mistral-common/.venv/bin/pytest`
- `PYTHONPATH=src /Users/julien.denize/workspace/mistral-common/.venv/bin/mypy .`
- Ruff checks on changed files

PR #295, opened 2026-08-25
Add agnostic validation mode

juliendenize:
## Summary
- Add `ValidationMode.agnostic`: a pipeline-agnostic validation mode that applies all content checks without a pipeline-specific final-role constraint
- Keep `continue_final_message` assistant-only and make it a no-op for other accepted roles
- Document final-message and continuation behavior for every validation mode
- Rebased on main (resolves conflicts with the docstring overhaul in #317)
- Renamed `structural` → `agnostic` per review feedback: the mode is the loosest (no terminal-role rule), so a "structure"-themed name was misleading
- Updated stale docstrings that still listed only `serving`, `finetuning`, `test` as available modes

## Testing
- `pytest tests/` (1542 passed, 17 skipped)
- `pytest --doctest-modules ./src` for validator.py and messages.py
- `mypy src/mistral_common/protocol/instruct/validator.py`
- `ruff check .` and `ruff format --check .`
- Brute-forced ~120 conversation/version combos confirming: agnostic raises only where both `test` and `finetuning` raise (the mid-conversation invalid `FinetuningAssistantMessage.weight` case raises the deepest content-level error, which is intended)

Stacked on #293.


docs/usage/experimental.md
```diff
@@ -51,7 +51,7 @@ mistral_common serve mistralai/Magistral-Small-2507 [validation_mode] \
 #### Command Line Options
 
 - `tokenizer_path`: Path to the tokenizer. Can be a HuggingFace model ID or a local path.
-- `validation_mode`: Validation mode to use, choices in: "test", "finetuning", "serving" (Optional, defaults to `"test"`)
+- `validation_mode`: Validation mode to use, choices in: "test", "finetuning", "serving", "general" (Optional, defaults to `"test"`)
```
joa-stdn: I think general is not good naming IMO, appendable or anything that says that it's not finished maybe?

docs/usage/requests.md
```diff
@@ -45,6 +45,19 @@ A conversation with a model is a sequence of messages. Each message can be a use
 
 Tools are functions that the model can call to get information to answer the user's question. See the [Tools](./tools.md) section for more information.
 
+### Continuing the final message
+
+Set `continue_final_message=True` on a [ChatCompletionRequest][mistral_common.protocol.instruct.request.ChatCompletionRequest] to have the tokenizer omit the closing special token of an unprefixed final assistant message, so the model continues writing it instead of starting a new turn. For a final user, system, or tool message, encoding is a no-op: tokens, text, multimodal outputs, and `prefix_ids` are unchanged. Whether validation accepts that request depends on the validation mode.
```
joa-stdn: not super cool that this is not compatible with our API protocol... do you know how we ended up with this?

tests/validation/test_chat_validation.py
```diff
@@ -118,6 +118,52 @@ def validator_v15(request: pytest.FixtureRequest) -> MistralRequestValidator:
 
 
 class TestChatValidation:
+    def test_general_accepts_unprefixed_final_assistant_without_continuation(self) -> None:
```
joa-stdn: can we parametrize these better?

src/mistral_common/tokens/tokenizers/instruct.py
```diff
@@ -197,12 +189,11 @@ def encode_instruct(
...
                 audios.extend(new_audios)
             elif isinstance(msg, AssistantMessage):
-                continue_message = request.continue_final_message and (msg_idx == len(request.messages) - 1)
-
+                continue_message = request.continue_final_message and is_final
                 new_tokens = self.encode_assistant_message(
                     msg, msg_idx < last_user_idx, continue_message=continue_message
                 )
-                if msg_idx == len(request.messages) - 1:
+                if is_final:
```
joa-stdn: I don't understand why prefix_ids is set even when continue final message is False? To me this sounds super weird but maybe I missed something?

src/mistral_common/protocol/instruct/validator.py
```diff
@@ -333,6 +339,12 @@ def _validate_message_order(self, messages: list[UATS]) -> None:
     def _validate_last_message(self, message: UATS, continue_final_message: bool) -> None:
         # The last message must be a user or tool message in serving mode or an assistant message in finetuning mode
         last_message_role = message.role
+        if self._mode == ValidationMode.general:
+            if continue_final_message and isinstance(message, AssistantMessage) and message.prefix:
```
joa-stdn: Actually I think it's super confusing that we're exposing here both continue_final_message and message.prefix?

joa-stdn (approved):

PR #296, opened 2026-08-26
Improve typing at dynamic tokenizer boundaries

juliendenize:
## Summary

- preserve concrete request types through generic normalizers without broad mypy suppressions
- describe the optional SentencePiece API through a local protocol
- narrow and document the remaining Pydantic generic-boundary ignore

## Compatibility

No public runtime signatures, request fields, tokenizer behavior, or `Tokenized` fields change.

## Validation

- `../../.venv/bin/mypy src/mistral_common tests`
- `../../.venv/bin/ruff check src tests`
- `../../.venv/bin/ruff format --check src tests`
- `../../.venv/bin/pytest` (1625 passed, 16 skipped)

PR #299, opened 2026-08-27
Align tokenizer type contracts

juliendenize:
## Summary

- reduce `InstructTokenizer` from four generic parameters to three
- align its abstract user/system encoding signatures with calls already made by `encode_instruct`
- distinguish `FieldBuilder` input and output types through an explicit conversion hook

## Stack

Depends on #296. Review this PR against `typing-boundary-cleanup`.

## Compatibility

- Bare `InstructTokenizer` usage remains source-compatible and behavior-preserving.
- Explicit four-parameter `InstructTokenizer[...]` annotations must drop the assistant-message parameter.
- `encode_user_message` implementers must accept optional `settings`; tokenizer implementations must provide `encode_system_message`. These methods were already called by `InstructTokenizerBase.encode_instruct`; the abstract interface now exposes that requirement directly.
- Direct `FieldBuilder[T]` subclasses must migrate to `FieldBuilder[InputT, OutputT]` and implement `_convert`; identity builders return their input unchanged.
- Assistant/Chat message annotation updates are annotation-only and behavior-preserving.
- `ChatCompletionRequest[...]`, `InstructRequest[...]`, runtime tokenizer outputs, and all `Tokenized` fields remain unchanged.

The compatibility inventory covered public source subclasses, test annotations, documentation examples, and exported symbols. No tracked four-parameter tokenizer usage remained. No cast or `TYPE_CHECKING`-only declaration hides the tokenizer interface.

## Validation

- `MYPYPATH=$PWD/src ../../.venv/bin/mypy src/mistral_common tests`
- `../../.venv/bin/ruff check src tests`
- `../../.venv/bin/ruff format --check src tests`
- `PYTHONPATH=$PWD/src ../../.venv/bin/pytest` (1633 passed, 16 skipped)

### DIFF
--- a/src/mistral_common/tokens/tokenizers/base.py
+++ b/src/mistral_common/tokens/tokenizers/base.py
@@ -12,10 +12,10 @@ from mistral_common.deprecation import warn_once
 from mistral_common.protocol.fim.request import FIMRequest
 from mistral_common.protocol.instruct.chunk import ContentChunk
 from mistral_common.protocol.instruct.messages import (
-    AssistantMessageType,
+    SystemMessage,
     UserMessage,
 )
-from mistral_common.protocol.instruct.request import InstructRequest
+from mistral_common.protocol.instruct.request import InstructRequest, ModelSettings
 from mistral_common.protocol.instruct.tool_calls import Tool
 from mistral_common.protocol.speech.request import SpeechRequest
 from mistral_common.protocol.transcription.request import TranscriptionRequest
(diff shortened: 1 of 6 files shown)


src/mistral_common/tokens/tokenizers/base.py
```diff
@@ -431,6 +428,7 @@ def encode_user_message(
         is_first: bool,
         system_prompt: str | None = None,
         force_img_first: bool = False,
+        settings: ModelSettings = ModelSettings.none(),
```
joa-stdn, line 431: not a fan of this, is this again for backward compatibility? This method is fully private and should not be used from the outside

joa-stdn (approved):

src/mistral_common/tokens/tokenizers/base.py
juliendenize: this is because InstructTokenizer implements this interface and to realign expected attributes.

PR #305, opened 2026-09-04
fix: allow v11 system prompts with audio

WangXukang-cypher:
## Summary

Allow the V11 instruct tokenizer path to accept system prompts when audio content is present.

This keeps the V7 behavior unchanged for audio requests while extending audio detection to include `AudioURLChunk`, and adds regression coverage for both inline audio chunks and audio URL chunks.

## Test Plan

- `python -m py_compile src\mistral_common\tokens\tokenizers\instruct.py tests\test_tokenizer_v11.py tests\test_tokenizer_v7_audio.py`
- `git diff --check`
- `PRE_COMMIT_HOME=.pre-commit-cache python -m pre_commit run --all-files --show-diff-on-failure`
  - Lint GitHub Actions workflow files: Passed
  - ruff check: Passed
  - ruff format: Passed
  - uv-lock: Passed

### DIFF
--- a/src/mistral_common/tokens/tokenizers/instruct.py
+++ b/src/mistral_common/tokens/tokenizers/instruct.py
@@ -1077,7 +1077,7 @@ class InstructTokenizerV7(InstructTokenizerV3):
         return any(
             isinstance(message, UserMessage)
             and isinstance(message.content, list)
-            and any(isinstance(chunk, AudioChunk) for chunk in message.content)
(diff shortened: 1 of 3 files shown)


juliendenize:
Hey thanks for the contribution however this has been made on purpose as we didn't release a model using tokenizer v11 and audio.

WangXukang-cypher:
So it's like this, I understand now. Thank you for your reply and inspection～～

PR #294, opened 2026-08-25
fix(tokenizer): validate ImageConfig fields to prevent ZeroDivisionError

shoemoney:
Fixes cryptic ZeroDivisionError when ImageConfig is constructed with invalid zero values.

**Bug:** `ImageConfig(image_patch_size=0)` raises `ZeroDivisionError` deep in `ImageEncoder._image_to_num_tokens` at `width_tokens = (w-1)//(image_patch_size*spatial_merge_size)+1` (image.py:97). Similarly `spatial_merge_size=0` and `max_image_size=0` give cryptic division errors far from the misconfiguration.

**Fix:** Add `__post_init__` to `ImageConfig` mirroring `AudioConfig.__post_init__` hardening in #253/#292. Validates `image_patch_size > 0`, `max_image_size > 0`, `spatial_merge_size > 0` with descriptive messages, failing fast at construction.

**Evidence:** Verified RED->GREEN - before fix `ImageConfig(0,1024,1)._image_to_num_tokens` raised `ZeroDivisionError: integer division or modulo by zero`, after fix it raises `AssertionError: image_patch_size must be > 0, got 0`. All 28 tests in `tests/test_image.py` pass.

Written in conjunction with my pair programmer Claude.

PR #259, opened 2026-07-06
Handle audio data URL MIME variants

fus3r:
## Summary

Audio data URL prefix stripping only matched word-only MIME subtypes. Valid data URLs such as `data:audio/x-wav;base64,...`, `data:audio/vnd.wave;base64,...`, and `data:audio/wav;codec=pcm;base64,...` were left untouched.

That made audio parsing try to base64-decode the full data URL rather than the payload, so valid audio bytes failed with a base64 or audio-format error. This affected both `AudioChunk.to_openai()` and tokenizer/audio parsing through `Audio.from_base64()` / `Audio.from_audio_chunk()`. The same bytes worked when the MIME subtype was simply `audio/wav`.

This broadens the audio data URL prefix match to allow MIME subtype punctuation and media-type parameters before `;base64,`, and makes `Audio.from_base64()` reuse the same helper as `AudioChunk.to_openai()`. The parameter match avoids consuming semicolon separators, so malformed inputs with repeated semicolons do not trigger pathological backtracking. The payload and detected format are unchanged.

## Testing

- `pytest tests/test_converters.py tests/test_audio.py -q`
- `ruff check src/mistral_common/protocol/instruct/chunk.py src/mistral_common/tokens/tokenizers/audio.py tests/test_converters.py tests/test_audio.py`
- `ruff format src/mistral_common/protocol/instruct/chunk.py src/mistral_common/tokens/tokenizers/audio.py tests/test_converters.py tests/test_audio.py --check`
- `mypy src tests scripts`
- `pytest tests/ --ignore=tests/integrations -q`
- `pytest --doctest-modules ./src -q`
- `mkdocs build --strict`
- `pytest tests/integrations/ -q`

No related issue. This follows the same audio conversion surface as #245.

PR #308, opened 2026-09-07
Fix image URL loading crashes and add a download timeout

### LINKED ISSUE #307: [BUG: `image_from_chunk` crashes on malformed data URLs, opens bare "file*" names as local paths, and downloads with no timeout]
### Python -VV

```shell
Python 3.12.14 (main, Aug 25 2026, 14:00:49) [Clang 22.1.3 ]
```

### Pip Freeze

```shell
mistral-common @ main (1fdcf24)
numpy==2.2.6
opencv-python-headless==4.11.0.86
pillow==11.3.0
pydantic==2.11.7
requests==2.32.4
tiktoken==0.12.0
```

### Reproduction Steps

`image_from_chunk` (`src/mistral_common/tokens/tokenizers/image.py`) sits on the encode path for every `ImageURLChunk`: `ImageEncoder.__call__` -> `InstructTokenizerV3._encode_content_chunk` -> `MistralTokenizer.encode_chat_completion`. Three inputs misbehave:

1. Data URL without a comma raises `IndexError` instead of a library error:

```python
from mistral_common.protocol.instruct.chunk import ImageURLChunk
from mistral_common.tokens.tokenizers.image import image_from_chunk

image_from_chunk(ImageURLChunk(image_url="data:image/png;base64"))
# IndexError: list index out of range   (image.py: data = chunk.get_url().split(",")[1])
```

2. The local-file branch matches `startswith("file")`, not `file://`. Any URL whose text starts with "file" is opened relative to the server's cwd:

```python
# with a file named file.png in cwd:
image_from_chunk(ImageURLChunk(image_url="file.png"))
# <PIL.PngImagePlugin.PngImageFile ... size=8x8>  (opened from disk, handle never closed)
```

3. `download_image` (`src/mistral_common/image.py`) calls `requests.get(url, headers=headers)` with no `timeout`. A server that accepts the connection and never responds blocks the encode call forever (Bandit B113).

### Expected Behavior

1. A malformed data URL should raise the same `RuntimeError` family the function already uses for unsupported schemes, not `IndexError`.
2. Only `file://` URIs should be treated as local files. A bare name like `file.png` should fall through to "Unsupported image url scheme".
3. `requests.get` should have a finite timeout so one dead URL cannot hang an encode worker.

### Additional Context

Reachable with untrusted input: `ImageURLChunk` is accepted on user messages, and the experimental tokenize server (`mistral_common.experimental.app`) turns the `IndexError` into an unhandled 500 since its handlers only catch `ValueError`. vLLM and the Transformers `MistralCommonBackend` go through the same `encode_chat_completion` path.

`Audio.from_url` in `src/mistral_common/tokens/tokenizers/audio.py` has the same missing timeout; I left it out to keep the fix reviewable in one function.

### Suggested Solutions

Split the data URL on the first comma and raise `RuntimeError` when there is no payload, require the `file://` prefix for the local-file branch (and close the handle), and pass a default `timeout` to `requests.get` in `download_image`. I have a branch with this fix plus tests that fail on `main` and pass with the change; PR to follow.

BarneyChambers:
## Summary

`image_from_chunk` mishandles three of the four URL shapes it accepts, and the HTTP branch can hang forever.

```python
if chunk.get_url().startswith("data:image"):
    data = chunk.get_url().split(",")[1]          # IndexError when there is no comma
    ...
if chunk.get_url().startswith("file"):            # matches "file.png", not just file://
    return Image.open(open(chunk.get_url().replace("file://", ""), "rb"))  # handle never closed
if chunk.get_url().startswith("http"):
    return download_image(chunk.get_url())        # requests.get with no timeout
```

Deterministic:

```
data:image/png;base64,<payload>   -> works              (unchanged)
data:image/png;base64             -> IndexError         (should be RuntimeError)
data:image/png;base64,            -> UnidentifiedImageError deep in PIL (should be RuntimeError)
file.png                          -> opens ./file.png   (should be unsupported scheme)
file:///tmp/ok.png                -> works              (unchanged)
https://<hanging-server>/x.png    -> blocks forever     (should time out)
```

This function is on the public encode path: `ImageEncoder.__call__` -> `InstructTokenizerV3._encode_content_chunk` -> `MistralTokenizer.encode_chat_completion`, which is what vLLM and the Transformers `MistralCommonBackend` call. `ImageURLChunk` content comes from the request, so all three inputs are attacker-supplied.

Fixes mistralai/mistral-common#307

## Real world example

A serving stack runs the experimental tokenize server (`mistral_common.experimental.app`) or any framework that calls `encode_chat_completion` on user requests. A client sends:

```json
{"messages": [{"role": "user", "content": [
  {"type": "image_url", "image_url": {"url": "data:image/png;base64"}}
]}]}
```

**What happens today:**

| Step | Result |
| --- | --- |
| `image_from_chunk` splits on "," | `IndexError: list index out of range` |
| Experimental app handlers | only catch `ValueError`; `IndexError` escapes as an unhandled 500 |
| Same request with `"url": "https://attacker.example/slow"` | `requests.get` never returns; the encode worker is gone until restart |

The `startswith("file")` arm is quieter: a URL like `file.png` is opened relative to the server's working directory, so a name collision reads a local file instead of raising "Unsupported image url scheme". The handle is also never closed.

**After this fix:** the malformed data URL and the bare `file.png` raise `RuntimeError` (the same family the function already uses for unsupported schemes), and the HTTP branch gives up after 10 seconds with the existing "Error downloading the image" wrapping, since `requests.exceptions.Timeout` is a `RequestException`.

## Fix

- Split the data URL on the first comma with `partition`; raise `RuntimeError` when there is no payload (covers both the missing comma and the empty payload).
- Require the `file://` prefix for the local-file branch and open the file in a `with` block (`Image.load()` before close), so bare names fall through to the unsupported-scheme error and the handle is closed. Behavior note: multi-frame images (GIF/TIFF) can no longer be seeked past frame 0 after return; the encoder only ever used frame 0, and the old code merely leaked the fd that made seeking possible.
- Pass `timeout` (default 10.0s) from `download_image` to `requests.get`. Exposed as a parameter so callers can tune it; existing error wrapping is unchanged.

## Not a duplicate

#289 and #294 fixed crashes in image sizing/config after loading; this PR is about loading itself. #259 touches audio data URL prefixes, different file. `Audio.from_url` has the same missing timeout but a different grammar, so it stays out of scope here; flagged in the issue as follow-up.

## Test

No existing test covered malformed data URLs, bare file names, or the timeout (the two existing mocks accept any call signature). Added:

- `test_image_from_chunk_data_url_without_payload` (both `;base64` and `;base64,`)
- `test_image_from_chunk_bare_file_name_is_unsupported`
- `test_image_from_chunk_file_uri` (regression guard for real `file://` URIs)
- `test_download_image_passes_timeout` (asserts `timeout=` reaches `requests.get`, and that a `Timeout` becomes `RuntimeError`)

The first, second and fourth fail on `main` and pass with this change:

```
uv run pytest tests/test_image.py -q
33 passed in 1.07s
```

Full unit suite: `uv run pytest tests/ --ignore=tests/integrations --ignore=tests/integration -n 4 --dist loadfile` -> 1233 passed, 16 skipped. Doctests, ruff check, ruff format and mypy all pass on the changed files.

## Verification

- Run the tests: see above (33 passed; 4 of the new asserts red on `main`)
- **Verify** the thing does what it should: malformed data URLs and bare file names raise `RuntimeError`; `requests.get` receives `timeout=10.0`
- **Verify** the thing does not do what it should not: valid data URLs, `file://` URIs and mocked HTTP downloads still encode byte-identically (existing `test_download_image`, `test_image_encoder_formats` untouched apart from mock signatures accepting the new kwarg)
- Supporting configuration: N/A; no configuration changed
- **Document**: docstring for `download_image` updated with the new `timeout` arg; no other user-facing API change


src/mistral_common/image.py
```diff
@@ -9,19 +9,20 @@
 from mistral_common import __version__
 
 
-def download_image(url: str) -> Image.Image:
+def download_image(url: str, timeout: float = 10.0) -> Image.Image:
```
juliendenize: Can we not make it a default here ?

src/mistral_common/tokens/tokenizers/image.py
```diff
@@ -64,16 +64,22 @@ def image_from_chunk(chunk: ImageURLChunk | ImageChunk) -> SerializableImage:
...
-        return Image.open(open(chunk.get_url().replace("file://", ""), "rb"))
-    if chunk.get_url().startswith("http"):
-        return download_image(chunk.get_url())
+    if url.startswith("file://"):
+        with open(url.removeprefix("file://"), "rb") as file:
+            image = Image.open(file)
+            image.load()
+        return image
+    if url.startswith("http"):
+        return download_image(url=url)
```
juliendenize: let's put timeout explicitly here. Also I think it should be an environment variable configurable default to ensure some heavy image pipelines can be configured

juliendenize (commented):
Thanks for your contribution it makes a lot of sense.

My concern is that I think timeout should be configurable by the environment with a nice error message to inform user they can increase it.

BarneyChambers:
> Thanks for your contribution it makes a lot of sense.
> 
> My concern is that I think timeout should be configurable by the environment with a nice error message to inform user they can increase it.

Thanks so much for your review @juliendenize. I think your recommendations make sense, I have implemented them in the most recent commit. 

`download_image` no longer defaults to `timeout=10.0`. Unset reads `MISTRAL_COMMON_IMAGE_DOWNLOAD_TIMEOUT` (10s if missing), and `image_from_chunk` passes that in explicitly. 

On timeout the error is:
timed out after {n} seconds. Pass a larger `timeout` or set the environment variable `MISTRAL_COMMON_IMAGE_DOWNLOAD_TIMEOUT` to increase the timeout.

So a heavy image pipeline can raise the cap without threading timeout through the tokenizer.

tests/test_image.py
```diff
@@ -19,6 +21,7 @@
     ImageConfig,
     ImageEncoder,
     SpecialImageIDs,
+	  image_from_chunk
```
juliendenize:
```suggestion
	image_from_chunk
```

juliendenize (changes requested):
Small follow-up to our earlier discussion: `image_from_chunk` resolves the timeout and `download_image` resolves it again when `timeout=None` — two resolution paths for one value. These suggestions keep the resolution in exactly one place (`download_image`'s `timeout is None` branch) and update the test to guard that `image_from_chunk` doesn't override the default resolution.

Also flagging: the latest commit (dcc54b7) leaves `image_from_chunk` in the `tests/test_image.py` import block with a tab instead of spaces and no trailing comma, which is a syntax error — `pytest tests/test_image.py` fails at collection on this branch. Suggestion attached.

Verified locally on top of this branch: `pytest tests/test_image.py` → 51 passed; `ruff check` + `ruff format --check` clean.

src/mistral_common/tokens/tokenizers/image.py
```diff
@@ -7,7 +7,7 @@
 import numpy as np
 from PIL import Image
 
-from mistral_common.image import SerializableImage, download_image
+from mistral_common.image import SerializableImage, download_image, image_download_timeout
```
juliendenize:
```suggestion
from mistral_common.image import SerializableImage, download_image
```

Becomes unused once `image_from_chunk` no longer resolves the timeout itself (see suggestion below).

src/mistral_common/tokens/tokenizers/image.py
```diff
@@ -71,16 +71,22 @@ def image_from_chunk(chunk: ImageURLChunk | ImageChunk) -> SerializableImage:
...
-        return Image.open(open(chunk.get_url().replace("file://", ""), "rb"))
-    if chunk.get_url().startswith("http"):
-        return download_image(chunk.get_url())
+    if url.startswith("file://"):
+        with open(url.removeprefix("file://"), "rb") as file:
+            image = Image.open(file)
+            image.load()
+        return image
+    if url.startswith("http"):
+        return download_image(url=url, timeout=image_download_timeout())
```
juliendenize:
```suggestion
        return download_image(url=url)
```

`download_image` already resolves the env var when `timeout=None` (`src/mistral_common/image.py`), so passing it here resolves the value twice. Keeping the default resolution in exactly one place avoids drift between the two paths.

tests/test_image.py
```diff
@@ -19,6 +21,7 @@
     ImageConfig,
     ImageEncoder,
     SpecialImageIDs,
+	image_from_chunk
```
juliendenize:
```suggestion
    image_from_chunk,
```

Heads-up: this line currently has a tab instead of spaces and no trailing comma (introduced in dcc54b7), which is a Python syntax error — `pytest tests/test_image.py` fails at collection and `ruff check` reports `Expected ',', found name`.

tests/test_image.py
```diff
@@ -115,6 +118,108 @@ def mock_get(url: str, headers: Any = None) -> Any:
...
+
+@pytest.mark.parametrize("timeout", [0.0, -1.0, float("inf"), float("nan")])
+def test_download_image_rejects_invalid_timeout(timeout: float) -> None:
+    with patch("mistral_common.image.requests.get") as mock_get:
+        with pytest.raises(RuntimeError, match="expected a positive finite number of seconds"):
+            download_image(url="https://example.com/image.png", timeout=timeout)
+    mock_get.assert_not_called()
+
+
+def test_image_from_chunk_http_passes_env_timeout(monkeypatch: pytest.MonkeyPatch) -> None:
```
juliendenize:
```suggestion
def test_image_from_chunk_http_uses_default_timeout(monkeypatch: pytest.MonkeyPatch) -> None:
```

With the change above, `image_from_chunk` no longer passes an explicit `timeout`, so the test should assert it delegates to `download_image`'s default resolution instead of asserting the env value is forwarded.

tests/test_image.py
```diff
@@ -115,6 +118,108 @@ def mock_get(url: str, headers: Any = None) -> Any:
...
+    monkeypatch.setenv(IMAGE_DOWNLOAD_TIMEOUT_ENV, "30")
+
+    with patch(
+        "mistral_common.tokens.tokenizers.image.download_image",
+        return_value=_create_test_image((8, 8)),
+    ) as mock_download:
+        image_from_chunk(chunk=ImageURLChunk(image_url="https://example.com/image.png"))
+    mock_download.assert_called_once()
+    assert mock_download.call_args.kwargs["url"] == "https://example.com/image.png"
+    assert mock_download.call_args.kwargs["timeout"] == 30.0
```
juliendenize:
```suggestion
```

juliendenize (changes requested):
Small follow-up to our earlier discussion: `image_from_chunk` resolves the timeout and `download_image` resolves it again when `timeout=None` — two resolution paths for one value. These suggestions keep the resolution in exactly one place (`download_image`'s `timeout is None` branch) and update the test to guard that `image_from_chunk` doesn't override the default resolution.

Also flagging: the latest commit (dcc54b7) leaves `image_from_chunk` in the `tests/test_image.py` import block with a tab instead of spaces and no trailing comma, which is a syntax error — `pytest tests/test_image.py` fails at collection on this branch. Suggestion attached.

Verified locally on top of this branch: `pytest tests/test_image.py` → 51 passed; `ruff check` + `ruff format --check` clean.

src/mistral_common/image.py
```diff
@@ -8,26 +10,64 @@
...
+    try:
+        timeout = float(raw)
+    except ValueError:
+        raise RuntimeError(
+            f"Invalid {IMAGE_DOWNLOAD_TIMEOUT_ENV}={raw!r}: expected a positive finite number of seconds."
+        ) from None
+    return _require_positive_finite_timeout(timeout, display=f"{IMAGE_DOWNLOAD_TIMEOUT_ENV}={raw!r}")
+
+
+def download_image(url: str, timeout: float | None = None) -> Image.Image:
```
juliendenize:
```suggestion
def download_image(url: str, timeout: float | None) -> Image.Image:
```

src/mistral_common/image.py
```diff
@@ -8,26 +10,64 @@
 
 from mistral_common import __version__
 
+IMAGE_DOWNLOAD_TIMEOUT_ENV = "MISTRAL_COMMON_IMAGE_DOWNLOAD_TIMEOUT"
```
juliendenize:
```suggestion
_IMAGE_DOWNLOAD_TIMEOUT_ENV_KEY = "MISTRAL_COMMON_IMAGE_DOWNLOAD_TIMEOUT"
```

src/mistral_common/image.py
```diff
@@ -8,26 +10,64 @@
...
+        raise RuntimeError(f"Invalid {display}: expected a positive finite number of seconds.")
+    return timeout
+
+
+def image_download_timeout() -> float:
+    r"""Return the HTTP timeout used when downloading images.
+
+    Reads `MISTRAL_COMMON_IMAGE_DOWNLOAD_TIMEOUT` (seconds). Unset or empty uses 10.
+    """
+    raw = os.getenv(IMAGE_DOWNLOAD_TIMEOUT_ENV)
```
juliendenize:
```suggestion
    raw = os.getenv(_IMAGE_DOWNLOAD_TIMEOUT_ENV_KEY)
```

juliendenize:
Hmm sorry about the comment it's generated obviously, i'm just fixing some stuff related to style / minor refactoring and then i'll merge. Thanks a lot for iterating and the contrib !

juliendenize (approved):

BarneyChambers:
> Hmm sorry about the comment it's generated obviously, i'm just fixing some stuff related to style / minor refactoring and then i'll merge. Thanks a lot for iterating and the contrib !

No problem at all, thanks for merging!

PR #312, opened 2026-09-13
Fix HTML-escaped JSON in tool call arguments for v2/v3/v7 templates

### LINKED ISSUE #311: Dict tool-call arguments render as HTML-escaped JSON in v2/v3/v7 chat templates
### Python -VV

```
Python 3.12.14 (main, Sep  1 2026, 00:10:07) [GCC 14.2.0]
```

### Pip Freeze

```
-e git+https://github.com/mistralai/mistral-common.git@e224216f4cdf1f3f3ff1484dbe648ef203e4e1cd#egg=mistral_common
Jinja2==3.1.6
MarkupSafe==3.0.3
```

### Reproduction Steps

Generate a v2/v3/v7 chat template and render a tool call whose `function.arguments` is a dict rather than a pre-serialized string:

```python
from jinja2.sandbox import ImmutableSandboxedEnvironment
from mistral_common.integrations.chat_templates.chat_templates import generate_chat_template
from mistral_common.tokens.tokenizers.base import TokenizerVersion

template = generate_chat_template(
    spm=False, tokenizer_version=TokenizerVersion.v3, image_support=False, audio_support=False,
    thinking_support=False, default_system_prompt=None, plain_thinking_support=False,
    use_special_token_variables=False,
)
env = ImmutableSandboxedEnvironment(trim_blocks=True, lstrip_blocks=True, keep_trailing_newline=True)
out = env.from_string(template).render(
    messages=[
        {"role": "user", "content": "weather?"},
        {"role": "assistant", "content": None, "tool_calls": [
            {"id": "123456789", "function": {"name": "get_weather", "arguments": {"city": "Paris"}}}
        ]},
    ],
    bos_token="<s>", eos_token="</s>",
)
print(out)
```

Output:

```
<s>[INST]weather?[/INST][TOOL_CALLS][{&#34;name&#34;: &#34;get_weather&#34;, &#34;arguments&#34;: {"city": "Paris"}, &#34;id&#34;: &#34;123456789&#34;}]</s>
```

### Expected Behavior

```
<s>[INST]weather?[/INST][TOOL_CALLS][{"name": "get_weather", "arguments": {"city": "Paris"}, "id": "123456789"}]</s>
```

The `"` around `name`/`arguments`/`id` come out as the HTML entity `&#34;` instead of a literal quote, so the segment isn't valid JSON. The inner `{"city": "Paris"}` itself is unaffected, only the hand-written punctuation around it.

### Additional Context

`_emit_argument_normalization` does `{%- set arguments = arguments|tojson|safe %}` when `arguments` isn't already a string. `tojson` returns a `markupsafe.Markup` (a `str` subclass), and the template then concatenates it into the JSON via plain `+`, e.g. `'{"name": "' + name + '", "arguments": ' + arguments + '}'`. Since `Markup` overrides `__radd__`, Python gives it priority over `str.__add__`, so `arguments.__radd__(...)` runs first and HTML-escapes the literal on its left, not `arguments`. Only v2/v3/v7 show it visibly; v11/v13/v15 use a `name[ARGS]arguments` format with no quotes next to `arguments`, so the same thing happens there with nothing to corrupt.

### Suggested Solutions

Jinja's `string` filter doesn't fix it (it's `soft_str`, which deliberately preserves `Markup`). `~ ''` does, since `~` compiles to `str_join`, which plain-`str()`s each piece before joining. Have a fix and regression test ready to open as a PR against this issue.

AmirF194:
Fixes #311.

`tojson` returns a `markupsafe.Markup` (a `str` subclass), and `Markup.__radd__` takes priority over `str.__add__` on a plain-string left operand. `_emit_argument_normalization` marked the normalized `arguments` `|safe`, so every place that later did `'{"name": "' + name + ...' + arguments + '}'` got the literal quotes around `name`/`arguments`/`id` HTML-escaped to `&#34;`, corrupting the JSON. The `arguments` value itself was never touched, which is why it was easy to miss in a spot check: only the surrounding hand-written punctuation broke.

Jinja's `string` filter doesn't fix this (it's `soft_str`, which deliberately preserves `Markup` so `escape()` still runs on it later). `~ ''` does, since `~` compiles to `str_join`, which plain-`str()`s every piece before joining, same as the reproduction in the issue confirmed by hand.

Added `test_tool_call_dict_arguments_produce_valid_json` (v2/v3/v3-spm/v7) asserting no `&#34;` in the output and that both `"name"` and `"arguments"` render as literal JSON. Reverting just the `~ ''` line turns it red again, same container both times. Regenerated the 26 golden `.jinja` fixtures that embed `_emit_argument_normalization`'s output; each diff is the single changed line. `pytest tests/`, `--doctest-modules ./src`, `ruff`, `mypy` and `pre-commit` are all clean, and the full `tests/integrations/` suite (parity + transformers) passes.


AmirF194:
Flagging this one, green for about a week with no review yet. The fix is one call in template_generator.py (swap the Markup concatenation for a plain str), the rest of the diff is the jinja fixture regeneration.

PR #313, opened 2026-09-14
test(tokens): add unit test suite for token sequence padding and truncation invariants

gcoinstash-cmd:
### Summary
Adds unit tests verifying token ID sequence truncation and pad length calculations:
- Confirms max-length truncation retains prefix tokens deterministically.
- Asserts boundary validation when sequence length matches exact max limit.


juliendenize:
Hey, i think you forgot to include some files in your PR

PR #321, opened 2026-09-17
fix(image): preserve float32 dtype and avoid float64 promotion in normalize (#316)

### LINKED ISSUE #316: [BUG] Image normalization promotes float32 inputs to float64
## Description

  `mistral_common.tokens.tokenizers.image.transform_image()` explicitly creates a `float32` NumPy array, but `normalize()` silently promotes it to `float64`.

  This doubles the memory used by processed images and significantly increases CPU preprocessing time. This surfaces downstream as a major Pixtral image preprocessing bottleneck.

  ### Reproducer

  ```python
  import numpy as np

  from mistral_common.tokens.tokenizers.image import (
      DATASET_MEAN,
      DATASET_STD,
      normalize,
  )

  image = np.zeros((1024, 1024, 3), dtype=np.float32)
  normalized = normalize(image, DATASET_MEAN, DATASET_STD)

  print(image.dtype)       # float32
  print(normalized.dtype)  # float64
```

  The promotion occurs [here](https://github.com/mistralai/mistral-common/blob/main/src/mistral_common/tokens/tokenizers/image.py#L145):

```
  np_image = (np_image - mean) / std
```

  mean and std are tuples of Python floats, so NumPy promotes the arithmetic result to float64.

  transform_image() initially requests the desired dtype explicitly:

```
  np_image = cv2.resize(
      np.array(_convert_to_rgb(image), dtype=np.float32),
      new_size,
      interpolation=cv2.INTER_CUBIC,
  )
```

  but returns a float64 array after normalization.

  ### Expected behavior

  Image normalization should preserve the float32 dtype established by transform_image(). For example:

```
  mean_array = np.asarray(mean, dtype=np_image.dtype)
  std_array = np.asarray(std, dtype=np_image.dtype)

  np_image -= mean_array
  np_image /= std_array
```
In-place operations used for further performance improvement.

### Performance impact
I benchmarked normalization on a batch of 16 preallocated 1024×1024 float32 images. Using the median of seven runs:

- Current: 275.8 ms 
- Proposed: 173.9 ms

That’s about a 1.6× speedup for normalization itself. This only covers normalization and tensor conversion—not decoding, resizing, or PIL-to-NumPy conversion—so end-to-end gains will depend on the workload and hardware.

It also cuts the intermediate memory use from 384 MiB to 192 MiB for this batch.

  ### Numerical impact

  For random 1024×1024 images that did not require resizing, the resulting BF16 tensors were bit-identical.

  In a separate test involving bicubic resizing, 319 out of approximately 2.18 million BF16 values differed (~0.015%). The maximum absolute difference was 0.0078125, one BF16 step at that magnitude.

  ### Environment

  - mistral-common: 1.11.7
  - Python: 3.12.12
  - NumPy: 2.3.5
  - OpenCV headless: 5.0.0.93
  - Pillow: 12.3.0
  - PyTorch: 2.13.0+cu130
  - Platform: Linux x86_64

### PR DESCRIPTION
**Pull request by andriiorap**
## Description

Fixes #316

In `mistral_common.tokens.tokenizers.image.transform_image()`, an image is explicitly created with `dtype=np.float32`. However, `normalize()` was performing `np_image = (np_image - mean) / std` with `mean` and `std` as Python float tuples. In NumPy, operating on a `float32` array with standard Python float scalars/tuples silently promotes the output array to `float64`.

This doubles intermediate image memory allocation during preprocessing and increases CPU normalization time.

## Proposed Changes

- In `normalize()` (`src/mistral_common/tokens/tokenizers/image.py`):
  - Convert `mean` and `std` to NumPy arrays matching `np_image.dtype` via `np.asarray(..., dtype=np_image.dtype)`.
  - Perform arithmetic in-place (`np_image -= mean_array`, `np_image /= std_array`). Because `np_image` was just created by `np_image / 255.0` on the line prior, in-place operations are safe and avoid two additional full-sized array allocations.
- In `tests/test_image.py`:
  - Added `test_normalize_dtype_preservation` asserting `normalize()` preserves `float32` and `float64` dtypes.
  - Added `test_transform_image_returns_float32` verifying `transform_image()` outputs `float32`.

## Performance & Memory Impact

Benchmarked on a batch of 16 1024×1024 float32 images:
- **Memory**: Halved intermediate memory usage from 384 MiB to 192 MiB.
- **Speed**: ~1.30x speedup in normalization runtime.
- **Numerical precision**: Bit-identical within float32 tolerance (`rtol=1e-6, atol=1e-6`).

## Checks
- [x] Ruff linter and formatter passed (`ruff check`, `ruff format`)
- [x] Mypy type checker passed (`mypy`)
- [x] Unit tests passed (`pytest tests/test_image.py tests/test_tokenizer_v3_mm.py`)

### CODE REVIEW

File: tests/test_image.py [hunk 1]
```diff
@@ -248,3 +248,39 @@ def mock_get(url: str, headers: Any = None) -> Any:
...
+    normalized = normalize(image, DATASET_MEAN, DATASET_STD)
+
+    # Verify input array was not modified in-place
+    np.testing.assert_array_equal(image, image_copy)
+
+    # Reference calculation
+    ref = (image_copy / 255.0 - np.array(DATASET_MEAN, dtype=np.float32)) / np.array(DATASET_STD, dtype=np.float32)
+    ref = ref.transpose(2, 0, 1)
+
+    np.testing.assert_allclose(normalized, ref, rtol=1e-6, atol=1e-6)
```
juliendenize: Not needed
```suggestion
```

tests/test_image.py
```diff
@@ -248,3 +248,39 @@ def mock_get(url: str, headers: Any = None) -> Any:
     for output in outputs[1:]:
         assert (output.image == outputs[0].image).all()
         assert output.tokens == outputs[0].tokens
+
+
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+def test_normalize_dtype_preservation(dtype: type) -> None:
+    from mistral_common.tokens.tokenizers.image import DATASET_MEAN, DATASET_STD, normalize
```
juliendenize: move import to the top of the file please

tests/test_image.py
```diff
@@ -248,3 +248,39 @@ def mock_get(url: str, headers: Any = None) -> Any:
...
+
+    image: np.ndarray = np.zeros((64, 64, 3), dtype=dtype)
+    normalized = normalize(image, DATASET_MEAN, DATASET_STD)
+    assert normalized.dtype == dtype, f"Expected {dtype} but got {normalized.dtype}"
+
+
+def test_transform_image_returns_float32() -> None:
+    from mistral_common.tokens.tokenizers.image import transform_image
```
juliendenize: move import to the top of the file please

juliendenize (changes requested):
Thanks for the contribution !

Could you address the comments i added to move import to the top of the file and delete an unnecessary test ?

andriiorap:
Thanks for the review @juliendenize! I have moved all imports to the top of the file in `tests/test_image.py` and removed the redundant precision test as suggested.

tests/test_image.py
andriiorap: Thanks, fixed!

tests/test_image.py
andriiorap: Done!

tests/test_image.py
andriiorap: Done!

juliendenize (approved):
thanks for iterating LGTM !

src/mistral_common/tokens/tokenizers/image.py
```diff
@@ -159,45 +127,39 @@ def normalize(
...
 
     Returns:
-        Normalized image with shape (C, H, W), scaled to unit variance.
-
-    Raises:
-        AssertionError: If `np_image` is not (H, W, C) or the channel count does
-            not match mean and std.
+        Normalized image with shape (C, H, W).
     """
     np_image = np_image / 255.0
```
juliendenize:
```suggestion
    np_image = np.divide(np_image, np.float32(255.0), dtype=np_image.dtype)
```

juliendenize (commented):
Just adding a quick suggestion i'll make vibe make needed changes to make it work the tests need some recalibration

src/mistral_common/tokens/tokenizers/image.py
juliendenize: Applied the suggestion (it's now on the PR head via f52e93d) and recalibrated `EXP_IMG_SUM` for the float32 pipeline.

Root cause of the failing sums: with `normalize()` computing in float32, the test's `np.abs(image).sum()` accumulates in float32, and that result varies across numpy versions (2.2.6 for py<3.13 vs 2.4.2 for py>=3.13, both pinned in `uv.lock`) by up to 0.375 — beyond the 1e-1 tolerance. Calibrating the constants alone can't satisfy both numpy versions.

Fix: accumulate the test sum in float64 (`sum(dtype=np.float64)`, bit-identical across environments) and recalibrate `EXP_IMG_SUM` against the pinned CI env (opencv 4.11.0.86).

Verified in CI on all matrix versions (3.10–3.14, both numpy pins) via validation PR #328 — all `build_lint_test` and `integration_tests` jobs pass. The recalibration commit is b8fe328 on `vibe/image-calib-321` (same tree as the PR head plus the test fix); I couldn't push directly to the fork branch from this environment.

andriiorap:
Pulled and merged the recalibration and float32 divide changes from `vibe/image-calib-321` into the branch. All CI checks are green now! Thanks @juliendenize!

PR #323, opened 2026-09-17
Validate tool results by call ID for v3 and v7

### LINKED ISSUE #322: V3 and V7 validators accept tool results with mismatched call IDs
## Description

`MistralRequestValidatorV3` and `MistralRequestValidatorV5` validate the number and ordering of tool calls and tool results, but they do not verify that each `ToolMessage.tool_call_id` references a pending `ToolCall.id`.

For example, this conversation is currently accepted in serving mode even though the result references a different call:

```python
messages = [
    UserMessage(content="Run the tool"),
    AssistantMessage(
        tool_calls=[
            ToolCall(
                id="123456789",
                function=FunctionCall(name="example", arguments="{}"),
            )
        ]
    ),
    ToolMessage(content="result", tool_call_id="999999999"),
]
```

This affects tokenizer versions v3 and v7. The v11+ validator already performs ID-aware pairing and rejects the same structure with `InvalidMessageStructureException`.

Expected behavior: once tool-call IDs are supported, a tool result should only satisfy the pending call with the same ID. Mismatched, duplicate, and unexpected result IDs should be rejected, while result ordering for parallel calls may remain flexible.

A focused fix can move the existing ID-aware pairing validation to the v3 validator so v3, v7, and newer validators share the behavior without changing v1/v2 compatibility.

shubhangi013:
## Summary

Fixes #322.

The v3 and v7 validators previously balanced tool calls and results by count only, so a `ToolMessage` could satisfy a different pending `ToolCall` as long as the counts matched. Tokenizer v11 and newer already validated this relationship by ID.

## Changes

- move the existing ID-aware tool call/result pairing validation to the v3 validator so v3, v7, and newer versions share it
- preserve v1/v2 behavior, where tool-call IDs are not required
- add focused v3/v7 regression coverage for matching and mismatched IDs
- update legacy v7 and finetuning fixtures so their tool call/result IDs are structurally coherent

## Verification

- `pytest tests/unit/mistral_common/protocol/instruct/test_validator.py -q` (4 passed)
- `pytest tests/validation/test_chat_validation.py tests/validation/test_finetuning_validation.py -q` (140 passed, 3 skipped)
- targeted tokenizer and finetuning regressions (2 passed, 1 skipped)
- `ruff check .`
- `ruff format --check .`
- `mypy src/mistral_common/protocol/instruct/validator.py`

- [x] Tests added or updated for changed behavior
- [x] Documentation and public API examples updated where needed
- [x] Ruff, mypy, and relevant pytest checks pass


juliendenize:
Hey this is normal behavior and expected.

